PR #6, opened 2016-01-13
Added operator overloading to ableC

krame505:
I have added operator overloading to ableC.  Let me know what you think.  

I basically added a Maybe dispatch attribute to type for each unary operator.   If the dispatch is not defined, then the usual production is used.  

For the binary operators, the double-dispatch resolution was inspired by operator overloading in Python.  There is a left and right dispatch attribute for each binary operator, and also an inherited attribute that provides the type of the other operand.  The left dispatch is checked first, if it is missing, the right is tried.  If there is no right dispatch, then the existing machinery for built-in types is used.  This also means that there is no additional error checking needed when overloading an operator; if you don't define it, then you get the error message that is from using invalid types in C.  

Many operators are also often defined in terms of each other.  For example, it would be tedious to write +=, *=, !=, etc. all manually when =, !, +, *, ==, etc. are already defined.  To remedy this, I am auto-generating these kinds of operators as defaults if their components are overloaded.  Currently, I am only generating > from <= and < from >= and vice versa, but it would be better if all these could be generated from one equality and one comparison operator.  I haven't done this yet as the naive approach would result in re-evaluating expressions, but I could do this by generating a statement-expression that evaluates the operands to temporary variables.  This would also require the and/or operators to be defined.  Anyway, there is a lot more I could be doing with operator default generation, but I don't want to make it too complicated.  The way it is it works well without any unexpected behavior.  

To make all this more concrete, I created an example string extension, providing strings with nice operations built in, such as appending, equality, etc.  There are examples included.

### DIFF
--- a/edu.umn.cs.melt.ableC/abstractsyntax/Expr.sv
+++ b/edu.umn.cs.melt.ableC/abstractsyntax/Expr.sv
@@ -60,6 +60,15 @@ top::Expr ::= e::Expr
 abstract production unaryOpExpr
 top::Expr ::= op::UnaryOp  e::Expr
 {
+  top.globalDecls := e.globalDecls;
+  top.defs = e.defs;
+  top.freeVariables = e.freeVariables;
+  
+  forwards to getUnaryOverload(op, e.typerep)(e, top.location);
+}
+abstract production unaryOpExprDefault
+top::Expr ::= op::UnaryOp  e::Expr
+{
   top.pp = if op.preExpr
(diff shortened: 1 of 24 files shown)


krame505:
Also sorry that there are some commits rolled into the pull request that I maybe should have pushed directly - I was working for a while without internet and couldn't push them earlier, then couldn't figure out how to push only some of them.  The last 2 are the most significant.

PR #9, opened 2016-02-28
Added host functor attribute

krame505:
This should fix the flow problems with pp on the env.  Note that this requires the latest update to Silver in the pull request for functor attributes.

### DIFF
--- a/edu.umn.cs.melt.ableC/abstractsyntax/AsmConstruct.sv
+++ b/edu.umn.cs.melt.ableC/abstractsyntax/AsmConstruct.sv
@@ -1,9 +1,10 @@
 
-nonterminal AsmStatement with location, pp, env, returnType, freeVariables;
+nonterminal AsmStatement with location, pp, host<AsmStatement>, env, returnType, freeVariables;
 
 abstract production asmStatement
 a::AsmStatement ::= arg::AsmArgument
 {
+  propagate host;
   a.pp = concat( [ text("asm ("), arg.pp, text(")"), text(";") ] );
   a.freeVariables = arg.freeVariables;
 }
@@ -11,14 +12,16 @@ a::AsmStatement ::= arg::AsmArgument
 abstract production asmStatementTypeQual
 a::AsmStatement ::= tq::Qualifier arg::AsmArgument
 {
+  propagate host;
   a.pp = concat( [ text("asm "), tq.pp, text(" ("), arg.pp, text(")"), text(";") ] );
   a.freeVariables = arg.freeVariables;
 }
 
-nonterminal AsmArgument with location, pp, env, returnType, freeVariables;
+nonterminal AsmArgument with location, pp, host<AsmArgument>, env, returnType, freeVariables;
 abstract production asmArgument
 top::AsmArgument ::= s::String asmOps1::AsmOperands asmOps2::AsmOperands asmC::AsmClobbers
 {
+  propagate host;
   top.pp = concat( [ text(s) ]
              ++ (if asmOps1Exists || asmOps2Exists || asmClobExists then [text(": ")] else [ ])  
              ++ [asmOps1.pp]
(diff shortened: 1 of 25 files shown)


ericvanwyk:
Looks good.  But shouldn't we be using the new ``host`` attribute somewhere?  I would expect that the driver gets the host attribute and the gets the pp off of it.  Or are there examples of using ``host`` somewhere else?

krame505:
I am currently using it in the root to get the pp there.  It could go in
the driver, but then you need to specify all the inherited attributes
there.  Or would it be better to have a separate attribute 'hostPP' on root
which gets used in the driver, leaving the original root pp intact?

Lucas Kramer

ericvanwyk:
OK, I'll look there.  What inherited attributes does host need on Root?  I wouldn't think it needs any.  I'm running to the colloquium now but will look in an hour when it is over.

krame505:
Hmm... I guess it shouldn't need any for the pp, but I we getting flow
errors.  I will take a look.

ericvanwyk:
OK, let me know what you find.

One way to use the host attribute would be to let the driver get the host tree and ask for pp on it.  It may also check if there are errors on the host tree.  If there were none on the original tree there certainly should be none here.  This is not how we would want to use this, we need something like Ted's non-inteference to ensure that this situation never arises.  But it would be useful now as a very rudimentary sort of testing.

ericvanwyk:
Lucas, have you built the closure extension using the --clean and --warn-all flags?  With just --clean, Silver (in the feature/functorattrib) branch takes 36 seconds on coldpress.  When adding the --warn-all flag it carries on for over 4 minutes, at which point I killed it.  Is this working for you?

krame505:
I get the same result.  The flow analysis is just taking around 5 minutes
on my machine (maybe longer for you), but it does still work.  I was going
to discuss this with Ted...

Lucas Kramer

tedinski:
re: analysis taking a long time. Check its memory consumption. You might need to up -Xmx from the default... The general symptom of running out of memory with Silver code is things just taking seemingly forever.

krame505:
Are you sure that is case if it does finish?  It seems that it couldn't
finish if out of memory.

Lucas Kramer

ericvanwyk:
I think I just killed it too soon.  I'll set the memory as Ted suggests and
see what happens.

krame505:
By the way, I get the same behavior when building host, so it doesn't have
to do with the closure extension.

Lucas Kramer

krame505:
It appears that we are intentionally making pp depend on the env, via
hackUnused.  I will try to fix this.

ericvanwyk:
OK.  What are these hackUnused things for?  I don't recall seeing them
before.  The name certainly suggests an issue...  There should be a comment
in the explaining the intention behind them.

krame505:
They basically are a way of manually changing the flow dependencies, since
we don't have a nice way of doing this at the moment. Something else to add
to silver...

krame505:
It appears that pp depends on the env in parameterDecl.  Why is, Ted?

ericvanwyk:
OK. But then there should be a comment above these things that says so.
Can you add one - if these end up surviving this round of edits?

I don't think pp should depend on env.  And what is happening in the
ppAttribute function has a comment "TODO: fix flow dependencies"...  So
something odd is happening here.

Ted, any thought?  Thanks.

tedinski:
No idea. Possible Silver bug? I'll look, but I might not have time to fix it today.

I agree pp should not depend on env. Perhaps prepare a patch removing the use of env (and returnType) from these functions.

tedinski:
The use of env was introduced in this patch, which unhelpfully does not explain why:

https://github.com/melt-umn/ableC/commit/26dedf8bbe53cedf7f04ae82f48999beb1383b29

krame505:
Strange. I don't see where it is actually used. My plane is taking off
soon, so I will need to wait on this until tomorrow.

tedinski:
Nope, no Silver bug, I was just misread the graph output. I think I was just in error when I added these, or perhaps there was some erroneous intermediate state that used references (or perhaps still is and I'm not seeing where!)

I'd just remove env from those functions.

ericvanwyk:
Lucas, based on this please go ahead and pull them out.  They will live in
this branch for a bit, so we can recover them if need be.  I don't see
where env is used either.

Something to do on the plane! :)

Cheers,
Eric

krame505:
OK, sounds good. I will go ahead and merge when I can push my other flow
fixes.

ericvanwyk:
Sounds good.

krame505:
Another issue - host on Expr depends on the env, but we are no longer able to provide this to expressions nested inside Attributes.  Not sure how to resolve this?  
Do you know if these expressions are always constant?  If so, we could simply provide a dummy env.

ericvanwyk:
Why is it Attributes cannot compute pp without env?   Passing env down to
Attributes so that they can compute their host attribute is OK.

What about getting rid of the lists of Attribute and Attrib and writing
cons and nil productions for lists of these non-terminal types?  Lists like
this alway look like we are taking a shortcut and not doing things in an
attribute grammar way.

krame505:
Oh, I suppose I could still let env occur on those nonterminals and just not provide them when computing the pp.  Also it appears that right now host is doing nothing on those nonterminals.  I think the correct solution would be using special list nonterminals, I will do that.

ericvanwyk:
Sounds like the right thing to me.  Make it all look like an AST, decorate with env but don't let pp depend on it.

I think the lists as they are now are a remnant of the quick implementation inspired by Clang, but I am not sure of that.

Cheers,
Eric

tedinski:
Hm, Yeah I think introducing "Attributes" and "Attribs" list nonterminals will probably solve all these problems.

krame505:
I have attempted to fix this, however, the flow analysis is still reporting that the pp depends on env.  Looking in the flow graphs, some of the dependencies are in the hackUnused decls?  This doesn't make any sense, because I am giving them an explicit equation.

tedinski:
Here's something hopefully helpful:

    egrep "^\"[^/]*/[^/]*pps?\" ->" flow-deps-direct.dot | egrep -v "(pp|pps|qualname|isJust)\";$"

krame505:
This should now be fixed by 138e9ee09337fbbfa68aff8bc919c77b8151b386, merging

PR #11, opened 2016-03-05
Added revised overloading

krame505:
I have made major changes to the way overloading is implemented, in addition to adding overloading for member access and combined member access/function calls.  I have also updated the demo extensions to use the new features.


krame505:
Merging after changes discussed with Eric.  Overloading should be pretty much finished for the foreseeable future.

PR #12, opened 2016-03-20
Added new mechanism for lifting global decls based on a functor attribute transformation of the host tree

krame505:
The core of the changes are in Lifted.sv.  This version works and could be merged, but there are still a few things that should be resolved eventually so we may want to wait:
* Decide on which mechanism should be used for getting the global environment (both currently work)
* Decide on where the tree transformations belong.  Probably not in the driver, but maybe not in the root either?  
* globalDecls should become a monoid attribute, once these are implemented in Silver.  
* Consider storing globalDecls as Decls instead of [Pair<String Decl>].  Doing so would greatly simplify things.  But maybe we also want to be able to lift things up out of expressions to the nearest statement level?  This seems like it is covered by stmtExpr, but it would maybe simplify things.  If we do end up doing this, I would suggest a second monoid attribute working with the same lifted attribute.  
* Also figure out how to deal with duplicate globalDecls.  Currently we just remove duplicates at the consGlobalDecl production.  I think this should be an error, and we should just suggest that every global decl that is injected has an unique name made with genInt.  
* Currently, we are adding the defs from the global decls to the current scope in consGlobalDecl, this is somewhat of a bug, but I don't know how we can do this the right way efficiently.


krame505:
Eric, I kinda want to merge this ASAP, since it is needed for closures which are needed for rewriting.  The first two items I listed above are rather simple and really should be done before merging.  The third item requires getting monoid attributes shaped up, which I have some thoughts on.  I am OK with dealing with the other items later after we merge.

ericvanwyk:
Lucas,

OK, for the first item, pick one, but leave the other there, just commented
out.  For the second, do the transformation on the root and we can late
change this if there is a compelling reason.  I agree that we can address
the remainder after merging.

I don't think merging will break anything - but you should run some tests
to be sure.

Cheers,
Eric

krame505:
I have been running into some issues with moving the transformations from the driver to the root - do you have an issue with leaving it in the driver for now?

krame505:
I actually ended up moving it to a new production in a new nonterminal that wraps Root.  Everything seems to be working, should I go ahead and merge?

ericvanwyk:
Hi Lucas,

Sure go ahead and merge - but lets discuss this new production/nonterminal
tomorrow.  Of the top of my head I'm not convinced this is needed, but you
may have seen something that escaped me.

-Eric

PR #13, opened 2016-05-11
improve error message when gcc doesn't support C11

TravisCarlson:
Error message was "CPP call failed."
Had to grep for that message to see that there's a --show-cpp option.
Had to run again with --show-cpp to see the command:
	gcc -E -x c -D _POSIX_C_SOURCE -std=gnu1x -I .  "test.xc" > test.gen_cpp
Had to run that command to see the root error:
	cc1: error: unrecognized command line option "-std=gnu1x"

This commit updates the error message to be:
CPP call failed: gcc -E -x c -D _POSIX_C_SOURCE -std=gnu1x -I .  "test.xc" > test.gen_cpp

If that's too much info, maybe "CPP call failed: try --show-cpp for more info"?

It would be nice to see the "-std=gnu1x" error immediately, but I'm not sure how
to print that out in Silver.

### DIFF
--- a/edu.umn.cs.melt.ableC/drivers/parseAndPrint/Driver.sv
+++ b/edu.umn.cs.melt.ableC/drivers/parseAndPrint/Driver.sv
@@ -43,7 +43,7 @@ IOVal<Integer> ::= args::[String] ioIn::IO
           systemM(
           fullCppCmd);
         if mkCppFile != 0 then {
-          printM("CPP call failed.\n");
+          printM("CPP call failed: " ++ fullCppCmd ++ "\n");
           return 3;
         } else {
           text :: String <- readFileM(cppFileName);
--- a/edu.umn.cs.melt.ableC/drivers/parseAndPrint/Driver.sv
+++ b/edu.umn.cs.melt.ableC/drivers/parseAndPrint/Driver.sv
@@ -55,7 +55,7 @@ IOVal<Integer> ::= args::[String] ioIn::IO
   else if !isF.iovalue then
     ioval(print("File \"" ++ fileName ++ "\" not found.\n", isF.io), 1)
   else if mkCppFile.iovalue != 0 then
-    ioval(print("CPP call failed.\n", mkCppFile.io), 3)
+    ioval(print("CPP call failed: " ++ fullCppCmd ++ "\n", mkCppFile.io), 3)
   else if !result.parseSuccess then
     ioval(print(result.parseErrors ++ "\n", text.io), 2)
   else if containsBy(stringEq, "--show-ast", args) then
--- a/edu.umn.cs.melt.ableC/drivers/parseAndPrint/Driver.sv
+++ b/edu.umn.cs.melt.ableC/drivers/parseAndPrint/Driver.sv
@@ -55,7 +55,7 @@ IOVal<Integer> ::= args::[String] ioIn::IO
   else if !isF.iovalue then
     ioval(print("File \"" ++ fileName ++ "\" not found.\n", isF.io), 1)
   else if mkCppFile.iovalue != 0 then
-    ioval(print("CPP call failed: " ++ fullCppCmd ++ "\n", mkCppFile.io), 3)
+    ioval(print("CPP call failed.\n", mkCppFile.io), 3)
   else if !result.parseSuccess then
     ioval(print(result.parseErrors ++ "\n", text.io), 2)
   else if containsBy(stringEq, "--show-ast", args) then


krame505:
This version of the driver is going away, FYI. The new version using monads
is on a feature branch. Eric, any opposition to merging the monad driver
first?

krame505:
This is probably a good idea, though a better solution would be to just
print the stdout resulting of running the cpp call.

ericvanwyk:
Hi all,

1. Travis, see if you can get this to also print the error messages
generated by CPP when it fails.  Thus, something that show the CPP command
and also the errors should appear when it fails.

2. Lucas, yes we want to get the monad stuff into develop as soon as
possible.  I see no reason not to pull it in now so that the functionality
is there.  But we need to finish the documentation.  I also think there
should be a nice tutorial in silver/tutorials that shows how to use these -
maybe a couple of sample grammars.  At least one for I/O only and then
another that does something interesting with state.   So, I think you can
do a pull-request to merge the monads feature branch - but then please
address these other issues.

3. Ted, is this really the right CPP command?  What are the requirements
that we are placing in the C compiler?  Any specific version of GCC?

Cheers,
Eric

TravisCarlson:
Hi all,

>  1. Travis, see if you can get this to also print the error messages generated by CPP when it fails. Thus, something that show the CPP command and also the errors should appear when it fails.

I talked to Ted about this last week and he found that the Java methods that would make this easy have only existed since 1.7. And, as Lucas mentioned, this version of the Driver is going away. Still, I think looking into the details of this would be a good exercise to help me better understand Silver.

Ted, what is preventing us from moving to Java 7? What are the methods needed that have only existed since 1.7 (I'm sorry I don't remember exactly if you had said)?

The system call in ableC appears to trace down to Util.java:system() in silver.

In ableC/edu.umn.cs.melt.ableC/drivers/parseAndPrint/Driver.sv:
```
local mkCppFile :: IOVal<Integer> =
    system(
      fullCppCmd,  
      if containsBy(stringEq, "--show-cpp", args)
      then print("CPP command: " ++ fullCppCmd ++ "\n", isF.io)
      else isF.io);
```
In silver/grammars/core/IO.sv:
(documentation here: https://github.com/melt-umn/silver/blob/develop/documentation/concepts/ffi/index.md)
```
{--
 - Executes a shell command.  Specifically executes 'bash -c'.  So, not fully cross-platform.
 -
 - Avoid using this if possible.  If you need an IO action not present, request it, please.
 -
 - Access to command's output is not directly available, but it is run in a shell. You can
 - redirect to a file and read that.
 -
 - @param s  The string for the shell to execute.
 - @param i  The "before" world-state token.
 - @return  The exit value of the subprocess.
 -}
function system
IOVal<Integer> ::= s::String i::IO
{
  return error("Not Yet Implemented: system");
} foreign {
  "java" : return "new core.Pioval(%i%, common.Util.system(%s%.toString()))";
}
```

In silver/java/src/common/Util.java:
```java
public static int system(String sb) {
        try {
                String cmdstr[] = {"bash", "-c", sb}; // TODO: platform dependency!
                Process p = Runtime.getRuntime().exec(cmdstr);
                p.waitFor();
                return p.exitValue();
        } catch (Exception e) {
                throw new RuntimeException(e);
        }
}
```

I don't fully understand how the IO tokens used in silver fit together with the underlying Java code. Util.java includes a print() method further down that simply calls `System.out.print();`, which seems to suggest there's nothing strange happening with stdout within silver. In fact, ableC's Driver.sv uses this print() when given --show-cpp, which has no problem printing to stdout. The system() call is also passed an IO token and looks similar to print() other than returning IOVal<Boolean> rather than IO. However, when I modify the first line of system() to be `System.out.println("got here");`, I never see the message. Additionally, if I remove the redirection of the CPP command in ableC's Driver.sv so the command is no longer `gcc -E -x c -D _POSIX_C_SOURCE -std=gnu1x -I . "example1.xc" > example1.gen_cpp` but `gcc -E -x c -D _POSIX_C_SOURCE -std=gnu1x -I . "example1.xc"` or `gcc -E -x c -D _POSIX_C_SOURCE -std=gnu1x -I . "example1.xc" | tee example1.gen_cpp`, I don't see the contents of example1.gen_cpp printed to stdout. Is stdout being redirected somewhere? Is there something about IO silver that I'm (likely) misunderstanding?

https://github.com/melt-umn/silver/blob/develop/documentation/concepts/io/index.md
https://github.com/melt-umn/silver/blob/develop/documentation/concepts/monads/index.md

In silver/grammars/core/IO.sv:

```
{--
 - Displays a string on standard out. Newlines are NOT automatically added.
 -
 - @param s  The string to print.
 - @param i  The "before" world-state token.
 - @return  The "after" world-state token.
 -}
function print
IO ::= s::String i::IO
{
  return error("Not Yet Implemented: print");
} foreign { 
  "java" : return "common.Util.io(%i%, common.Util.print(%s%.toString()))";
} 
```

In silver/runtime/java/src/common/Util.java
```java
public static Object print(String sb) {
        // TODO: should we avoid demanding stringcatter objects?
        System.out.print(sb);
        return null;
}
```

```java
/**
 * This exists only because some things like {@link Util#writeFile} don't take an IO object, so
 * we use this to demand the io object, then call writeFile.
 * 
 * <p>In other cases, we're actually returning something, like system, and the call to the IOInteger
 * nonterminal constructor takes care of demanding the old IO object.
 * 
 * @param i First thing to do
 * @param o Second thing to do
 * @return null, the IO object.
 */
public static Object io(Object i, Object o) {
        return null;
}
```

Finally, some pointers on the development build process would be helpful, as would any documentation or description of how the generated jars fit together. Several times I modified a file, recompiled, ran my test again, and didn't notice any change. I ended up doing a deep-clean and rebuilding both silver and ableC from scratch after each modification to silver.

krame505:
Not sure I understand the confusion, or why we need Java 7. You should just
be able to redirect stderr to a file in the CPP command and read that.
As for your confusion with the silver build process, you need to manually
copy the silver.composed.Default.jar file from build to jars if you only
run self-compile for silver. You will need to rebuild ableC if you want to
test the new changes to silver.
On May 17, 2016 8:55 PM, "TravisCarlson" <notifications@github.com> wrote:

Hi all,


   1. Travis, see if you can get this to also print the error messages
   generated by CPP when it fails. Thus, something that show the CPP command
   and also the errors should appear when it fails.

I talked to Ted about this last week and he found that the Java methods
that would make this easy have only existed since 1.7. And, as Lucas
mentioned, this version of the Driver is going away. Still, I think looking
into the details of this would be a good exercise to help me better
understand Silver.

Ted, what is preventing us from moving to Java 7? What are the methods
needed that have only existed since 1.7 (I'm sorry I don't remember exactly
if you had said)?

The system call in ableC appears to trace down to Util.java:system() in
silver.

In ableC/edu.umn.cs.melt.ableC/drivers/parseAndPrint/Driver.sv:

local mkCppFile :: IOVal<Integer> =
    system(
      fullCppCmd,
      if containsBy(stringEq, "--show-cpp", args)
      then print("CPP command: " ++ fullCppCmd ++ "\n", isF.io)
      else isF.io);

In silver/grammars/core/IO.sv:
(documentation here:
https://github.com/melt-umn/silver/blob/develop/documentation/concepts/ffi/index.md
)

{--
 - Executes a shell command.  Specifically executes 'bash -c'.  So,
not fully cross-platform.
 -
 - Avoid using this if possible.  If you need an IO action not
present, request it, please.
 -
 - Access to command's output is not directly available, but it is run
in a shell. You can
 - redirect to a file and read that.
 -
 - @param s  The string for the shell to execute.
 - @param i  The "before" world-state token.
 - @return  The exit value of the subprocess.
 -}
function system
IOVal<Integer> ::= s::String i::IO
{
  return error("Not Yet Implemented: system");
} foreign {
  "java" : return "new core.Pioval(%i%, common.Util.system(%s%.toString()))";
}

In silver/java/src/common/Util.java:

public static int system(String sb) {
        try {
                String cmdstr[] = {"bash", "-c", sb}; // TODO:
platform dependency!
                Process p = Runtime.getRuntime().exec(cmdstr);
                p.waitFor();
                return p.exitValue();
        } catch (Exception e) {
                throw new RuntimeException(e);
        }
}

I don't fully understand how the IO tokens used in silver fit together with
the underlying Java code. Util.java includes a print() method further down
that simply calls System.out.print();, which seems to suggest there's
nothing strange happening with stdout within silver. In fact, ableC's
Driver.sv uses this print() when given --show-cpp, which has no problem
printing to stdout. The system() call is also passed an IO token and looks
similar to print() other than returning IOVal rather than IO. However, when
I modify the first line of system() to be System.out.println("got here");,
I never see the message. Additionally, if I remove the redirection of the
CPP command in ableC's Driver.sv so the command is no longer gcc -E -x c -D
_POSIX_C_SOURCE -std=gnu1x -I . "example1.xc" > example1.gen_cpp but gcc -E
-x c -D _POSIX_C_SOURCE -std=gnu1x -I . "example1.xc" or gcc -E -x c -D
_POSIX_C_SOURCE -std=gnu1x -I . "exampl e1.xc" | tee example1.gen_cpp, I
don't see the contents of example1.gen_cpp printed to stdout. Is stdout
being redirected somewhere? Is there something about IO silver that I'm
(likely) misunderstanding?

https://github.com/melt-umn/silver/blob/develop/documentation/concepts/io/index.md
https://github.com/melt-umn/silver/blob/develop/documentation/concepts/monads/index.md

In silver/grammars/core/IO.sv:

{--
 - Displays a string on standard out. Newlines are NOT automatically added.
 -
 - @param s  The string to print.
 - @param i  The "before" world-state token.
 - @return  The "after" world-state token.
 -}
function print
IO ::= s::String i::IO
{
  return error("Not Yet Implemented: print");
} foreign {
  "java" : return "common.Util.io(%i%, common.Util.print(%s%.toString()))";
}

In silver/runtime/java/src/common/Util.java

public static Object print(String sb) {
        // TODO: should we avoid demanding stringcatter objects?
        System.out.print(sb);
        return null;
}

/** * This exists only because some things like {@link Util#writeFile}
don't take an IO object, so * we use this to demand the io object,
then call writeFile. *  * <p>In other cases, we're actually returning
something, like system, and the call to the IOInteger * nonterminal
constructor takes care of demanding the old IO object. *  * @param i
First thing to do * @param o Second thing to do * @return null, the IO
object. */public static Object io(Object i, Object o) {
        return null;
}

Finally, some pointers on the development build process would be helpful,
as would any documentation or description of how the generated jars fit
together. Several times I modified a file, recompiled, ran my test again,
and didn't notice any change. I ended up doing a deep-clean and rebuilding
both silver and ableC from scratch after each modification to silver.

—
You are receiving this because you commented.

Reply to this email directly or view it on GitHub
<https://github.com/melt-umn/ableC/pull/13#issuecomment-219800151>

tedinski:
Ah. The `system` function should already print things out when a command it run, but there was a problem with Java less than 7 not having the methods to do this on ProcessBuilder.

IIRC, Jenkins is actually building using Java 7 now, and nobody noticed apparently.

Hey Eric, I think we should just commit to 7 as a minimum now. If you agree, I'll commit a fix to the Silver runtime that'll make `system` behave.

TravisCarlson:
Hi Lucas,

> You should just be able to redirect stderr to a file in the CPP command and read that.
I can successfully redirect the command's stderr to a file, but I would like to print the error message to stderr (preferably stderr, but at least stdout) so the user sees it. Does redirecting stderr to a file help me to do that? Are you saying that to print the output of a command to the screen that I need to redirect the output to a file, read the file, print the result, then remove the file?

> As for your confusion with the silver build process, you need to manually copy the silver.composed.Default.jar file from build to jars if you only run self-compile for silver. You will need to rebuild ableC if you want to test the new changes to silver.
That's very helpful. That manual copy after running self-compile is what I was missing. Is there not a script that both compiles and copies files where needed?

ericvanwyk:
Ted,  will you be in on Thursday?  It might be good for you and Travis to
sit down and have a conversation about this.  I think he could benefit from
some guidance on these specific issues - Java 7, etc.

-Eric

krame505:
Yes, that is what I meant. Maybe slightly hacky but it should work.
The reason self-compile doesn't do the copy is so that you have a chance to
test the new jars before nuking your current jars in case you broke
something major. Deep-rebuild checks that the new jars can be able to
recompile silver before overwriting the default ones.
Ted, is this more formally documented anywhere?

ericvanwyk:
Travis,

You can find some documentation here:
http://melt.cs.umn.edu/alpha/silver/doc/dev/building/   The source for this
documentation is in the Silver repo - in the "documentation" directory.
Please suggest (or better yet, make) changes to improve this.

It would be nice to find a less hacky solution to this problem that what
Lucas suggested (and admits is a hack).

Cheers,
Eric

krame505:
Yeah, the final solution would be to create a new io function that executes
a shell command and returns the stdout and stderr in a pair.

TravisCarlson:
That would be useful, yes, but I think it's beyond and not quite is needed here. The system call interface doesn't need to change. The implementation just needs to change to print the result of the command as a side effect. The ableC code wouldn't need to change. It would still be:

```
local mkCppFile :: IOVal<Integer> =
    system(
      fullCppCmd,  
      if containsBy(stringEq, "--show-cpp", args)
      then print("CPP command: " ++ fullCppCmd ++ "\n", isF.io)
      else isF.io);
```

I think what you're talking about, returning something that provides access to the output of the command, is more similar to what is usually called popen(). What is called system() usually prints the output without providing access to it and returns the exit code. Both are nice to have, but the latter is what we need here and is very similar to what we already have.

* Some examples of system()
https://docs.python.org/2/library/os.html#os.system
http://perldoc.perl.org/functions/system.html
http://linux.die.net/man/3/system

* Some examples of popen()
https://docs.python.org/2/library/os.html#os.popen
https://docs.python.org/2/library/subprocess.html#popen-constructor
http://perldoc.perl.org/functions/open.html
http://linux.die.net/man/3/popen

tedinski:
I forgot this was open. Is this change still needed/useful with the change to `system` or should we close this pull request?

krame505:
Would still be useful, but we can't merge now that the driver has been
rewritten to use monads.  Travis, if you want to go ahead and update this
with the current driver, that would be fine.

Lucas Kramer

TravisCarlson:
Sure, I'll make this same change to the current driver later today.

TravisCarlson:
I've made the same change again after reverting the original and merging with 'develop'.

Separately, it seems that the monadic driver has a bug where it prints the error message three times.

```
$ java -jar ableC.jar example1.xc 
CPP call failed: gcc -E -x c -D _POSIX_C_SOURCE -std=gnu1x -I .  "example1.xc" > example1.gen_cpp
CPP call failed: gcc -E -x c -D _POSIX_C_SOURCE -std=gnu1x -I .  "example1.xc" > example1.gen_cpp
CPP call failed: gcc -E -x c -D _POSIX_C_SOURCE -std=gnu1x -I .  "example1.xc" > example1.gen_cpp
$ java -jar ableC.jar undefined.xc
File "undefined.xc" not found.
File "undefined.xc" not found.
File "undefined.xc" not found.
```

krame505:
Yes, we know about that.  It is a bug with Silver's core IO system, Ted and
I are going to try to fix it today.

Lucas Kramer

krame505:
I guess you can go ahead and merge this if there are no other complaints.  
Btw, a temporary hack to fix IO monads being executed multiple times is to replace the last two lines of bindIO (in grammars/silver/core/monad/IO.sv) with 
```
  top.stateOut = stateOut;
  top.stateVal = stateVal;
```
and just don't commit that.  We can talk Thursday about fixing monads.

PR #15, opened 2016-06-03
Feature/monad driver

krame505:
Rewrite of driver using new support for monads.


krame505:
Eric OK'd merging this.

PR #18, opened 2016-07-08
Feature/misc bug fix

ericvanwyk:
I enabled the `prefix separator` defined in the host language.  This is used in the `using_transparent_prefix` examples in the `ableC_sample_projects` repo on http://github.umn.edu/melt.

I also changed this separator from ":" to "::".  I chose "::" becuase it is the same as the C++ scope operator and therefor may be more familiar to some users.  Previously this was ":", but the single colon is Silver's grammar separator and we should pick a prefix separator that is driven by syntax in object language (here, C and related language C++) and not on Silver's syntax.


krame505:
Sounds good.  The tests/examples in the algDataTypes extension should also get updated to make use of this.

ericvanwyk:
OK - I'll look at the algDataTypes stuff next.  Thanks for the pointer.

PR #21, opened 2017-02-10
Feature/templating

krame505:
I revised the vector extension to make use of the current lifting mechanism.  Productions that could benefit from templating to define generic functions in a header file are commented as such, with the ASTs currently written manually in local attributes.  I also made a few bug fixes to the string extension.  

This required a few changes to the host:
* Added global decl injection productions for Stmt, BaseTypeExpr, and Type
* Added globalDeclsEnv attribute to Decls to track what globalDecls were inserted by a previous consGlobalDecls production to avoid duplicates
* Added env check to remove globalDecls from being defined that are already in the env
* Added an attribute on types to perform name mangling
* Added '>' to the follow set of TypeName_c to fix MDA errors
* Implemented type checking on memberExpr


krame505:
Any reason not to merge this?  I will open a second pull request when I fix #22.

Issue #22, opened 2017-02-15
Extensions can forward to an AST containing errors in types that are uncaught


krame505:
When forwarding to directTypeExpr, there is no checking done to ensure that the parameter type is actually valid (e.g. for typedefType, that the parameter name is a valid type).  This can result in us generating non-compiling c code without raising errors.  This in theory shouldn't cause issues for the end user assuming extensions are bug-free, but can be an issue for the extension writer, as it points to a bug in the host language when there actually isn't.  

One way of fixing this may be to add attributes on Type to translate back to a BaseTypeExpr and TypeModifierExpr (similar to .lpp and .rpp), and have directTypeExpr forward to a production wrapping the result of this transformation instead.  This would also greatly simplify handling lifting globalDecls from Type, since then we would no longer be considering Type part of the bottom-level host AST.

krame505:
Fixed by #24.

Issue #23, opened 2017-02-18
BaseTypeExpr can contain TypeModiferExpr via directTypeExpr


krame505:
I found this issue when working on #22, as there is no obvious solution for what directTypeExpr(ty) should forward to when ty.typeModiferExpr is not baseTypeExpr().  

Currently, the issue shows up in situations like variableDecls, where there is a BaseTypeExpr and multiple TypeModiferExprs that are based on this type.  The pp of directTypeExpr is just the pp of type that it wraps, but this means that if you have variableDecls with multiple Declarators and the BaseTypeExpr being directTypeExpr(pointerType(...)), then the * is printed before the Declarator list, so only the first one ends up being a pointer.  

The fundamental issue here is that we shouldn't be able to have a TypeModifierExpr occur within a BaseTypeExpr in any way (at least in the host tree), so this is yet another reason why directTypeExpr really shouldn't be part of the host language.  But since doing so is useful, we need to have some method of performing lifting of TypeModifierExprs to the top of BaseTypeExpr, then lowering and inserting them into the bottom of TypeModifierExpr to replace baseTypeExpr.  

Here is my proposal to do that, which goes hand-in-hand with my other refactoring work to make directTypeExpr no longer part of the host:

* Add a production typeModiferTypeExpr (BaseTypeExpr ::= BaseTypeExpr TypeModifierExpr).  This is used for directTypeExpr(ty) to forward to when ty.typeModifierExpr is not baseTypeExpr()
* Add synthesized and inherited attributes to lift a list of TypeModifierExprs from BaseTypeExpr and send them down the TypeModiferExpr tree.  These must be provided from one to the other in every production that somehow wraps a BaseTypeExpr and one or more TypeModifierExprs (e.g. typeName, variableDecls, some extensions, etc.)  
* Define the host attribute on typeModiferTypeExpr to be its parameter BaseTypeExpr
* Define the host attribute on baseTypeExpr to be the new tree constructed from the parameter TypeModiferExprs

Since these changes also involve some further refactoring of the lifting stuff, I will be committing this in the feature/template branch, and will push once #21 is merged.

krame505:
Fixed by #24.

PR #24, opened 2017-02-21
Fix handling of directTypeExpr to make use of lifting

krame505:
This fixes issues #22 and #23.  

## TypeModiferExpr lifting
I added two new productions on BaseTypeExpr.  builtinTypeExpr takes a BuiltinType and wraps it as a BaseTypeExpr.  typeModifierTypeExpr takes a BaseTypeExpr and a TypeModifierExpr, wrapping them into a BaseTypeExpr.  This TypeModifierExpr must be moved to the bottom of the TypeModiferExpr tree corresponding to the BaseTypeExpr before code can be generated.  This is done via the lifted transformation.  

To do this, BaseTypeExpr has a new synthesized attribute `typeModifiers :: [TypeModiferExpr]`, and TypeModiferExpr a new inherited attribute `typeModifiersIn :: [TypeModiferExpr]`.  These collect all TypeModifierExprs in a BaseTypeExpr, in order of nesting, and send them back down the TypeModiferExpr tree to baseTypeExpr.  These attributes must be provided from BaseTypeExpr to TypeModifierExpr everywhere these are used (declarator, parameterDecl, etc.)  

The lifted attribute on typeModifierTypeExpr is then defined to just be the parameter BaseTypeExpr, and lifted on baseTypeExpr is defined to be the first TypeModifierExpr from typeModifiersIn, with its baseTypeExpr substituted for the next element of typeModifiersIn, etc.  This is computed by decorating the head of the list with typeModifiersIn as the tail of the list, then requesting lifted from that so that the process works recursively.  

## directTypeExpr translation
Types now have two new attributes, baseTypeExpr and typeModiferExpr, to compute the BaseTypeExpr and TypeModiferExpr for a Type.  This is fairly straightforward, except for the case of the gcc extension vectorType, since it required defining a BaseTypeExpr containing attributes.  We probably want to implement a production to do this, but that is something to fix at a later date.  

directTypeExpr can now forward to typeModifierTypeExpr with the type expressions that were created from the Type.  The pp and typerep are overridden to be the pp and parameter type, so that special types that extensions define are preserved in the original tree.  

This also nicely handles lifting on Type, since injectGlobalDeclsType can just turn into injectGlobalDeclsTypeExpr via this transformation.  Performing globalDecl lifting at the same time as the  transformation is not an issue, since the globalDecls and lifted both get accessed at the same time for BaseTypeExpr, and TypeModifierExpr gets redecorated with the same environment to access lifted once it gets transformed.  

## Lifting refactoring
I also did some refactoring of globalDecls lifting.  

I modified the injection productions to no longer forward to the lifted expression and instead define all attributes manually.  This removes a potential problem with interference.  

I also added a GlobalDecls nonterminal that mirrors Decls, but with a few slightly different attributes related to lifting.  This removes the forwarding dependency on globalDeclEnv, and also lets us get rid of a few equations that are `error("This shouldn't ever get accessed on Decls")` on the right-hand side.

### DIFF
--- a/edu.umn.cs.melt.ableC/abstractsyntax/Decls.sv
+++ b/edu.umn.cs.melt.ableC/abstractsyntax/Decls.sv
@@ -75,6 +75,7 @@ top::Decl ::= storage::[StorageClass]  attrs::[Attribute]  ty::BaseTypeExpr  dcl
   top.freeVariables = ty.freeVariables ++ dcls.freeVariables;
   
   dcls.baseType = ty.typerep;
+  dcls.typeModifiersIn = ty.typeModifiers;
   dcls.isTypedef = false;
   dcls.givenAttributes = attrs;
 }
@@ -101,6 +102,7 @@ top::Decl ::= attrs::[Attribute]  ty::BaseTypeExpr  dcls::Declarators
   top.freeVariables = ty.freeVariables ++ dcls.freeVariables;
   
   dcls.baseType = ty.typerep;
+  dcls.typeModifiersIn = ty.typeModifiers;
   dcls.isTypedef = true;
   dcls.givenAttributes = attrs;
 }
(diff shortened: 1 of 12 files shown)


tedinski:
Looks okay to me.

PR #26, opened 2017-02-28
change defs to be a collection attribute

TravisCarlson:
Change defs to be a collection attribute in preparation for an upcoming change to the cilk extension. We plan to create unique IDs for each scope and store the ID that each variable was declared in. 

@krame505 , before this is pulled, could you help to comment about how this relates to interference?


krame505:
Added a comment explaining the invariant on this.  

Something we should discuss: should the other 'defs' attributes (functiondefs, localdefs) also be collection attributes?  I am leaning towards being consistent on this, since there may be cases where an extension would want to define special things in these scopes as well.

TravisCarlson:
I've updated the other defs attributes as suggested by @krame505 and merged with the latest 'develop'. Any comments before this is ready to be merged?

PR #28, opened 2017-03-02
Improved resources for AST construction

krame505:
Same pull request as before, just on a different branch now.  

This consists of two more-or-less independent features that together allow us to do complex AST construction much more nicely.  

## Substitution mechanism
This allows for performing substitutions of declRefExpr for other Exprs, typedefTypeExpr for other BaseTypeExprs, Names for other Names, etc.  These are represented by a closed nonterminal Substitution, that is parameterized by the target name and item to substitute, such as `declRefExprSubstitution :: (Substitution ::= String Expr)`.  Each Substitution production has a corresponding Maybe attribute that looks at an incoming synthesized attribute containing a name to substitute and returns just what is to be substituted or nothing.  
To evaluate Substitutions, there are functions of the form `subExpr :: (Expr ::= [Substitution] Expr)` that take a list of Substitutions and perform them on an AST.  These fold the Substitution list into a Substitutions, which then are passed down the AST as an inherited attribute.  Every production has an aspect production in a separate grammar which defines the `substituted` functor attribute with the result of the substitution.  Every production that could be substituted decorates the incoming Substitutions with the name to substituted, and sets the substituted attribute according to the value of the appropriate attribute on Substitution.  

## Parsing wrapper functions
These are a group of functions that take a string of host code, call an parser wrapping the appropriate nonterminal, and either return an AST or error with parse errors.  These are of the form `parseExpr :: (Expr ::= String)`, and can function more-or-less as a drop-in replacement for txtExpr, txtStmt, etc.  For complex ASTs that wrap other ASTs, the embedded AST can be replaced by a name that is then replaced via the above substitution mechanism.  
One challenge with this is that in C, parsing depends on the list of type names that are in scope, which isn't set correctly for parsing fragments of code.  To fix this, for these parser wrappers specifically (but not exported for the host AST) a concrete production is added to add the syntax `proto_typedef foo;` to add a type name to the parser attribute for types.  
For performance reasons, commonly used ASTs should be declared as a global instead of within a production, to avoid calling the parser repeatedly.  This can however sometimes come at the cost of readability, requiring the use of substitutions instead of textually specifying pieces of code like names.  

## Future work
Instead of calling the parser with long strings of code, we could instead think of writing a Silver extension that would allow C code to be embedded directly in the Silver program, which would translate to the resulting AST.  This would fix the performance issues with calling the parser, and would also make things much more readable.

### DIFF
--- a/edu.umn.cs.melt.ableC/abstractsyntax/AsmConstruct.sv
+++ b/edu.umn.cs.melt.ableC/abstractsyntax/AsmConstruct.sv
@@ -1,10 +1,10 @@
 
-nonterminal AsmStatement with location, pp, host<AsmStatement>, lifted<AsmStatement>, substituted<AsmStatement>, env, substitutions, returnType, freeVariables;
+nonterminal AsmStatement with location, pp, host<AsmStatement>, lifted<AsmStatement>, env, returnType, freeVariables;
 
 abstract production asmStatement
 a::AsmStatement ::= arg::AsmArgument
 {
-  propagate host, lifted, substituted;
+  propagate host, lifted;
   a.pp = concat( [ text("asm ("), arg.pp, text(")"), text(";") ] );
   a.freeVariables = arg.freeVariables;
 }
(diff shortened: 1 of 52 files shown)


krame505:
Ted said this seems generally OK, merging.

PR #30, opened 2017-03-09
fix INTERNAL compiler error when using labelStmt

TravisCarlson:
When checking that a label has been declared precisely once in a function, lookup the label in any scope, not only the local scope.

Without this change, "INTERNAL compiler error: expected to find label in function scope, was missing" occurs when compiling the following:

```c
int main(void) {
l:
  return 0;
}
```


tedinski:
LGTM. I thought for a bit that maybe we did care about the scope it would look in, but normal lookup stops at the nearest scope that finds a result so this should be correct, afaict.

Issue #32, opened 2017-03-15
Incorrect handling of qualifiers for pointers in typeAssignableTo


krame505:
This *should* be a subsetting relation, but right now we are checking that all qualifiers are identical.  This means that something like
```
char *a;
const char *b = a;
```
is considered an error, while this should really be legal.  As a temporary workaround in the feature/templating branch, I am disabling checking of qualifiers on pointers completely.

PR #33, opened 2017-03-22
Revise lifting mechanism, implement template extension, utilize templates for vector extension

krame505:
Wow.  OK, so there are a lot of things that happened here.  In roughly chronological order:

## Redesign lifting mechanism to avoid redecoration/duplicate removal
This fixes a lot of performance issues, hacks, and general uglyness of the old system.  We also insert global decl defs into the correct scope when we perform the initial declaration of the host tree.
Instead of letting global decls be generated and decorated repeatedly and then removing duplicates at the top level, we ensure that decoration only happens once, in the first place where a global decl is generated, and put it in the environment at the global level.
Changes from the old way of doing things:
* The globalDecls attribute that passes the list of global decls up the tree now has type `[Decorated Decl]` instead of `[Pair<String Decl>]`, since we only do the decoration once and no longer need to keep track of duplicates.
* We want the env for everything to the right of a production to contain all the defs at a global level from global decls for things to its left.
  * defs on the new globalDecls from the head of consGlobalDecl are added to the environment for the tail of consGlobalDecl before being passed up the tree.
  * We have the invariant on defs/env that for each host production, its defs contains all the defs from all its children that do not open a new scope.  In addition, for each child of a host production, its env contains all defs from all children to its left that did not open a new scope.  Several bugs were found where this wasn't the case, and if you ever get errors about duplicate defs in the lifted tree, that is probably the cause.
  * Then at every place where we open a new scope, we need to somehow insert the defs from global decls for previous children at the global level.
  * To do this, we add a new type of Def, globalDefsDef, which wraps a list of defs to insert at the global level.  These are collected in Defs by another attribute, and in addEnv_i we insert their contribs into the the last (global) scopes for their respective namespaces.
  * In every host production, wherever the defs from a child do not get propagated up, that child's defs get wrapped in globalDefsDef and included in defs.
* It is now left to the extension writer to avoid introducting duplicate global decls.
  * Whenever an extension wants to introduce a global decl that may be needed in more than one place, the extension should introduce a Decl that first looks up what is being declared in the env, and forward to `decls(nilDecl())` if it already exists.
  * To avoid needing to write this kind of repetitive wrapper function, we introduce a new forwarding production, `production maybeDecl Decl ::= (Boolean ::= Decorated Env) Decl` which takes a function that is applied to the env to determine whether to forward to the provided Decl if true or to `decls(nilDecl())` if false.  

## Implement template extension
An extension that does C++-style templates was also implemented.  The purpose of this extension is:
* For end-users of ableC, to provide a general mechanism of type polymorphism
* For extension writers, to allow for extensions (e.g. vector) that utilize a large amount of static, type-dependant code to be able to write this in a header file
  * Note that extensions wanting to use templating for writing type-generic implementation code don't need to rely on templating for the extension type - an extension could define a type directly without relying on the template extension, but still write type-generic library code using the template extension.
  * Alternatively, an extension type could use the template extension to define a new type without any associated templated functions.  
* As a demonstration of the the use of lifting, substituion, and extension of the env with new namespaces.
This required allowing the addition of new namespaces to the environment, and allowing injecting arbitrary defs into the env.  This is done by makind Def a closed nonterminal with some additional rules imposed, and adding a new production defsDecl that injects new arbitrary defs into the AST.  
See [here](https://github.com/melt-umn/ableC/blob/feature/templating/extensions/templating/README.md) for examples and more info.  

## Modify vector extension to make use of template extension
The vector extension was re-structured to utilize the template extension instead of using the parsing/substitution/lifting mechanism explicitly.  This permitted a lot of code to move into header files, which removes overhead from multiple parser calls and simplifies the code considerably.
* Generated vector struct becomes a templated struct
* vectorTypeExpr forwards to `directTypeExpr(vectorType())`
* vectorType forwards to `injectGlobalDeclsType(templateTypeExprInstDecl(...), pointerType(structType(...)))`, where struct name and refId resulting from the instantiation are known by calling `templateMangledName`.  

## Further refactoring of closure extension
* Fix capturing semantics
  * Non-global functions can now be captured as function pointers, and arrays as pointers.
  * Everything captured is now const, since what is in the environment is a copy of what was captured.
* Update to use revised lifting mechanism
* Make closure type env-independant again
  * Allowed by updated lifting mechanism - can just forward to `injectGlobalDeclsType(maybeDecl(...), ...)`.
* Fix bugs with handling type qualifers
* Type expression syntax passes MDA

## Other misc. stuff
* Fix lots of bugs with handlings of defs/env
  * Invariant was broken that all defs from children that are not in a new scope should go in the parent defs.
* Workaround for bug #32, checking type compatability for pointers should make use of a subset relation, not checking that these are identical.
  * For now, just ignore pointers when checking type equality.  This needs to be fixed...

# Things to discuss before this is merged
* injectGlobalDeclsType is currently broken in that it is supposed to be a host production, but it forwards to get the right type-checking behavior.  Figure out a better solution for this.
* When should defsDecl get removed?  Seems like this should happen with host (since it is only used to determine forwarding behavior, and otherwise acts like nilDecls()) but that breaks the invariant that host should be propagated on all non-forwarding productions.  Currently it is removed by lifted.
* Making globalDecls have type `Decorated Decls` instead of `[Decorated Decl]` may be slightly more efficient, would avoid unfolding and folding anywhere we want to access defs.
* Additional features missing from template extension
  * Templated function calls with type inference?
  * Explicit global instantiations?  Would be easy to implement, just need concrete syntx for `templateExprInstDecl` or `templateTypeExprInstDecl`.
  * Allow linking behavior of templated functions to be explicitly specified, if desired.  Currently these are just static, to avoid redefinitions.  Maybe allow an option to emit duplicates and use the C++ linker?


krame505:
More TODO issues:
* Figure out recursive instantiations with different types in struct templates.  Generate forward declarations?  
* Add function to generate template struct refIds for other extensions to use.

krame505:
* Remove declsDef using host attribute instead of lifted, there doesn't really need to be an invariant on host about it being only always propagated.  
* Fix operator overloading to be non-interfering... doesn't *absolutely* need to happen before this is merged, but may be related.

krame505:
I've been looking at the bug that Nathan noticed at the meeting last week, where we have templated structs that contain an instantiation of themselves, resulting in ableC crashing in infinite recursion.  Generating the typedef with a struct prototype prior the struct fixes this; i.e. instead of
```
typedef struct __attribute__ (refId("...")) <name> {
  ...
} <name>;
```
generating
```
typedef struct <name> <name>;
struct __attribute__ (refId("...")) <name> {
  ...
};

This fixes that issue, but now we can't specify the refId in the first (tag) definition, because C only permits attributes to be defined where a struct is actually defined with a body, not on tag defs.  Using attributes to provide the refId was always a hack in the first place though, so maybe we should just go ahead and define custom abstract syntax for this.  
At this time I don't believe we are actually using this anywhere to specify refIds in concrete syntax, but we could re-add custom concrete syntax if this is really needed.  Worst-case, extensions that need to refer to a specific struct in a header file by refId could simply utilize the template extension.

krame505:
Fixed the above-mentioned issue with Ted's idea of using declsDef to insert a fake struct prototype with the refId we want.  This does mean we can't remove declsDef anymore when generating the host tree, since it will still be needed to ensure that the refId is consistent.

krame505:
I think I have things working without injectGlobalDeclsType - it turns out most of the reasons why this was needed can be fixed in other ways now that lifting avoids decorating duplicates.  However, there are may still be bugs in these extensions that aren't caught by any of the test cases.  
We do need to formalize invariants for how extensions with parametric type expressions should deal with parameter types that define something - something along the lines of that every type expression that is a child of an extension production is included in the forward tree exactly once, although `directTypeExpr(ty.typerep)` may occur any number of times.

krame505:
With that done, I think the only thing to do before merging is to figure out what to do about operator overloading.  Should we fix this now, or in a future feature branch?

ericvanwyk:
Let's put this in a future feature branch.  Maybe short-term future, but
create a branch for it.

Thanks - this is great work on all of this.

-Eric

krame505:
I believe this is ready to merge.  @tedinski @ericvanwyk @TravisCarlson any final comments?  I intend to merge on Thursday.

PR #34, opened 2017-04-01
feature/type_qualifiers

TravisCarlson:
This implements subtyping through type qualifiers in the model of Foster et al. "A theory of type qualifiers". Also adds some support in the host language for creating new, "pluggable" type qualifiers. Fixes some bugs in type checking, especially related to const.

This will conflict with feature/templating. I can wait until templating is merged, then fix the conflicts before merging this. I'm opening the pull request now because I'd appreciate any feedback.

See https://github.umn.edu/melt/edu.umn.cs.melt.exts.ableC.nonnull and https://github.umn.edu/melt/edu.umn.cs.melt.exts.ableC.enumerate for example uses of this.

Const is an example of a "positive" qualifier; `int *` is a subtype of `const int *` and so the former can be used wherever the latter is required, but not vice versa. In contrast, nonnull is an example of a "negative" qualifier; `int * nonnull` is a subtype of `int *`. Declaring `void foo(int * nonnull p)` then can ensure at compile time that `foo` is always passed a `p` qualified as nonnull. And since NULL is of type `void *`, not `void * nonnull`,  it cannot be assigned to any pointer qualified as nonnull.

## changes

* TypeQualifiers.sv
  * replace `qualCheck` attribute with `qualIsPositive`, `qualIsNegative`, and `qualAppliesWithRef`
  * create `pluggableQualifier` production
* ExprUnaryOps.sv
  * add `collectedTypeQualifiers` attribute to `UnaryOp`. In the nonnull extension, `aspect production addressOfOp` is created with `top.collectedTypeQualifiers <- [nonnullQualifier()];`. This attribute should probably  be added to any nonterminal with a `typerep` to support future extensions, but I didn't want to make too many changes that I didn't currently need. 
* TypeOps.sv
  * `compatibleTypes` and `typeAssignableTo` changed to treat qualified types as subtypes
* Types.sv
  * remove `withoutTypeQualifiers` attribute on `Type`, no longer used

## bug fixes

* Subtyping behavior of const: `void foo(const int *p)` can now be passed `int *`
* Disallow assignment to const: `const int n; n = 0;` is now an error
* Qualifiers treated as set: `const int *` is now the same as `const const int *`
* Dereferencing a non-pointer: `int x; *x;` is now an error
* Set errors in `parenExpr`: if `x = y` is an error then `(x = y)` should be too

I noticed this bug, but haven't fixed it: `char str[] = "abc";` shouldn't be an error


krame505:
Seems very interesting and useful!  A few comments:
* Qualifiers would seem to be more natural to become a closed nonterminal than having a single production in the host to which all new qualifiers must forward.  @tedinski do you have an opinion on this?  
*  In your nonnull qualifier example, do we really want to check for all dereference operations that they are non-null?  This would cause issues when working with an existing codebase, for example.  I would recommend adding new custom non-null dereference / deref-and-field-access operators that check this condition, instead of modifying the errors for the host operator.  
* Do we really want to get rid of `withoutTypeQualifiers`?  This seems useful for some extensions, although it may be possible to replace it instead with some attribute to remove only specific type qualifiers.  
* Conflicts with feature/templating shouldn't be too bad, and I don't know how long it will be before that is merged.  But we should probably test this with the various extensions changed in feature/templating before merging.  Let's decide what to do about that when one or both branches is ready to merge.  

I'll take a closer look at this tomorrow, I will let you know if I have more questions.

TravisCarlson:
> In your nonnull qualifier example, do we really want to check for all dereference operations that they are non-null? This would cause issues when working with an existing codebase, for example. I would recommend adding new custom non-null dereference / deref-and-field-access operators that check this condition, instead of modifying the errors for the host operator.

Great idea. Do you have any suggestion for the concrete syntax of the operators?

I haven't thought through all of the issues with type qualifiers as independently-developed, composable extensions, and this isn't the only issue.

For example, for a positive `tainted` qualifier -- that annotates strings coming from a possibly-malicious user -- to be useful, we would want to change the signature of `fgets()` to return `char * tainted`. Alternatively, we could use a negative `untainted` qualifier and use "safe" functions  (e.g. `int safe_printf(const char * untainted format, ...)` where desired, similar to the `safe_malloc()` used in the `nonnull` example, but we then run into problems if another extension tries to assign a value to a string qualified as `untainted`: even if a string definitely does not come from a user and is thus safe to assign to a variable of type `char * untainted`, how can that be handled in an extension that doesn't know about `untainted`? Eric gave the example of using `nonnull` together with the algebraic data types extension: if we write a program with a datatype qualified as `nonnull`, then a type error will occur on the adt-side's use of `malloc()`, which returns `void *` rather than the necessary `void * nonnull`.

> Do we really want to get rid of withoutTypeQualifiers? This seems useful for some extensions, although it may be possible to replace it instead with some attribute to remove only specific type qualifiers.

I can easily add it back in if you think it still might be useful. I was thinking about creating an attribute to remove specific type qualifiers, but was just trying to avoid adding too many things that don't have a use yet.

ericvanwyk:
Lucas, what is your motivation for wanting to add new deref operators
instead of using the existing one?  I'm not sure what you're getting at
with respect to "existing code base"?  I would think you just don't use the
extension on existing files - but do use it on new files.  Eventually one
may want to modify the code so that it is safe w.r.t. the non-null
qualifiers.

I think the extension is more interesting as one that provides guarantees
about non-null usage - even if it requires some work on the existing code
to make it compliant.

There is the unresolved issue of generated calls to malloc - is this what
you're referring to?  There is certainly some more work to do on this issue.

-Eric

krame505:
The issue with modifying the host operators is one of interference - if another extension (or, for that matter, library with a header containing static functions) performs operations that do not satisfy the added error checks, then that will introduce errors in code over which the user has no control.  In general, it is a 'bad thing' to allow extensions to modify the behavior of host constructs, unless we work out a very specific semantics for this as in the case of operator overloading.  
The only way around this would be to introduce some mechanism to automatically infer additional qualifiers not written explicitly as part of a type, but this would require major changes to the ableC type system.  
Calls to functions that lack the qualifiers in their result types, such as malloc, could be fixed by casting, but I am referring to places where the user has no ability to modify the code.  

It would be good to be able to implement a more sophisticated types of inference for non-null qualifiers, such as inferring that a variable is non-null after a conditional checking that is the case, but that would require modifying the type of something that is already in the environment.  We need to spend more time thinking about what this entails.

ericvanwyk:
Lucas,

Your comment was about an "existing code base" not interference.

I'm well aware that this violates Ted's notion of interference and that
modifying host language constructs is a 'bad thing' ... I have some
experience with this stuff... :)

For files that one controls, one can compile them using this extension.
For library file or other one that the programmer doesn't control of
doesn't want to change they can compile those by not using this extension.

But you are correct that we do need to think more about this and the type
system in general.  So there are lots of interesting problems yet to solve!

-Eric

remexre:
Just for the case of `malloc`, though, null can definitely be returned -- `malloc(0)` is called out in the manpage as a case where it will.

I think the issue of modifying `fgets()` to return a `tainted char*` is the bigger one, because that would definitely break static functions. (And more than a few libraries use static header functions, because lto still isn't widespread.)

It's not elegant, but one thing that'd work would be to only use the qualifier version of the stdlib functions in functions that use qualified types?

TravisCarlson:
> Just for the case of malloc, though, null can definitely be returned -- malloc(0) is called out in the manpage as a case where it will.

Suppose the writer of the ADT extension knows that a user might want to ensure that null is never assigned to certain variables, and is willing to add such checks within the extension. For example,  the ADT extension could use its own `safe_malloc()` that never returns null. Even then, the ADT extension can't independently annotate anything as nonnull, so can the error be avoided even in cases that we know are safe?

> I think the issue of modifying fgets() to return a tainted char* is the bigger one, because that would definitely break static functions. (And more than a few libraries use static header functions, because lto still isn't widespread.)

I don't understand why static functions are a problem, could you explain more? The use of type qualifiers is entirely on the ableC end; the qualifiers are simply dropped in the generated code that is used by gcc. I think my question is how we modify library functions in the first place, and how we handle two extensions that both want to modify the same function. Can we create aspect productions that check the names of functions being added to the environment, and add qualifiers to the types of certain known functions? Even if we can modify the qualified types of library functions, how do we handle the problems that will create for other extensions? e.g. Will other extensions then not be able to use`fgets()` because they can't annotate anything the result is assigned to as `tainted`?

> It's not elegant, but one thing that'd work would be to only use the qualifier version of the stdlib functions in functions that use qualified types?

That's what I was suggesting through the use of  `untainted` with `safe_printf()`. Then you only get safety guarantees where you remember to use the safe functions. This seems similar to the issue of whether a separate dereference operator should be created; do we want to ensure that all dereferences are safe and break the existing code base (or require a different compiler be used for it), or do we want to limit safety to these explicitly "safe" operators?

I'm not sure if this is what you were referring to, but alternatively the creation of a `tainted_fgets()` for use with `tainted` doesn't seem like such a good idea. It wouldn't be an error for the existing code base to assign `char *str = fgets(...);` but we would want that to be an error because `str` can then be used in places we require not to be `tainted`.

remexre:
`safe_malloc` is a better solution than modifying `malloc`, but returning NULL is probably the right choice for `malloc` with a size of zero. Alternately, we could return `malloc(1)` instead?

I was suggesting that if we do change the types of builtin functions, we only change them for functions that otherwise use the extension, so

```c
void foo(positive int x) {
    malloc(x);
}

void bar(int x) {
    malloc(x);
}
```

gets translated to

```c
void foo(int x) {
    safe_malloc(x);
}

void bar(int x) {
    malloc(x);
}
```

TravisCarlson:
Hi @krame505 and others,
Could you please take another look at this and comment on any changes you'd like to see before this is merged?

Thanks,
Travis

krame505:
Looks much better than before, but I still have a few (OK, a lot) comments:
* When you add qualifiers to a type, it appears that they are just appended without checking for duplicates.  So when type qualifiers are added via `collectedTypeQualifiers`, you may end up with duplicates?  Unless I'm not understanding what is going on here.  But addTypeQualifiers should probably avoid adding duplicate qualifiers anyway.  
* Since you closed `Qualifier`, `host` will need to work differently in order to eliminate extension `Qualifier`s in the host AST.  This can be done by writing a default production on `Qualifer`, which returns some sort of new null `Qualifer` production (`lifted` should just default to returning the `Qualifier` unchanged.)  But there is a better solution than this, which brings me to my next point:
* Currently, `host` and `lifted` aren't even doing anything on `Qualifier`, since propagate doesn't work across lists.  You could 'fix' this by manually rewriting all the equations on all productions that have `[Qualifier]` as a child to use `map`, but that would be a ton of work and very bug-prone.  The 'right' way of fixing this is to introduce a `Qualifiers` nonterminal, with `consQualifier` and `nilQualifier` productions.  This nonterminal could have an attribute to convert back to `[Qualifer]`, if that makes type checking easier.  Instead of having `host` occur on `Qualifer`, have an attribute `qualIsHost` (or something) that defaults to false, but is set true by the host productions.  Then when computing `host` on `Qualifiers`, this attribute can simply be checked and the qualifier possibly excluded from the list.  
* I don't understand the purpose of having `showlpp` and `showrpp` in addition to `lpp` and `rpp` on types.  This breaks all extensions with types that don't have these attributes defined, so all extension types will show up as what they forward to in error messages.  
* In fact, I don't think `qualname` is needed in the first place - just use pp?  It seems that the only other thing `qualname` is used for name-mangling qualifiers, but `qualname` doesn't seem to always be a valid C identifier, e.g. in enumerate.  I would recommend dropping `qualname` and just having mangledName occur on `Qualifier`.  
* On a related note, it appears that extension `Qualifier`s have an empty `pp`.  This shouldn't be needed, since extension qualifiers should be removed by the host, as mentioned previously.

krame505:
I just realized another issue with the 'nonnull' extension, in its current form where host constructs are modified.  This was only seeming to work due to the above-mentioned bug, where extension qualifiers were not being removed from the host.  As a result, the added host error checks should be failing in the host tree that lacks these extension qualifiers.  
There may be a way to fix this with some sort of additional collection attributes of analyses to perform, but figuring out a good way of doing this will take some thought.  For now my recommendation is to temporarily switch these extensions to use custom syntax, and if in the future we figure out a good way of making the host extensible in this way, then we can do that.  But if we want to get this merged soonish, then changing these for now would be the better option.

TravisCarlson:
Thank you Lucas for the very helpful comments.

> So when type qualifiers are added via collectedTypeQualifiers, you may end up with duplicates? 
That's right. I'm not bothering to avoid adding duplicate qualifiers because duplicates have no effect on the semantics. What is the reason you think duplicates should be avoided?

> The 'right' way of fixing this is to introduce a Qualifiers nonterminal, with consQualifier and nilQualifier productions.
Please take a look at the latest commits and let me know if you think I'm on the right track.

> I just realized another issue with the 'nonnull' extension, in its current form where host constructs are modified.
Yes, this is now an issue for the 'nonnull' extension. I'd prefer to work out a way to allow the analysis as-is than use custom syntax. Do you think this needs to be solved before merging? There are outstanding issues with the 'nonnull' extension other than this (e.g. silencing qualifier errors in code originating in extensions), and there will likely be more changes coming to the host language related to type qualifiers (e.g. flow-sensitive type qualifiers), that I'd like to submit a separate pull request later for. Keeping this branch up-to-date and fixing merge conflicts with 'develop' is getting to be a hassle. Currently the branch is useful enough for the 'enumerate', 'len_gte', and 'dimensionalAnalysis' extensions to be working fairly well, and various bugs related to the standard type qualifiers have been fixed, so I think there's some benefit in merging once the main issues have been worked out, even if more changes will come later.

Thanks,
Travis

ericvanwyk:
I don't see harm in merging this now.  Some issues remain but it does do
some cool things now.

krame505:
Duplicates should be avoided because of performance, and also printing an inferred type with a large number of repeated qualifiers (e.g. `nonnull nonnull nonull int*`) is suboptimal.  

Doesn't `len_gte` suffer from the same issue as `nonnull`?  But yes, I am fine with merging even if these two extensions are broken for now.  

`qualIsHost` looks good, but there should probably be a default production setting this to false, since all extension qualifiers shouldn't be considered host.

krame505:
Also, the `Qualifiers` bit looks good, but this does need to be updated as well for the closure, string, vector, and templating extensions.  I would do this, but I probably won't have time until next weekend.  This for sure shouldn't be merged until those are fixed, since the templating extension is being used for the go-concurrency extension, and it still would be bad to break the others.  
This refactoring might be a reasonable task for some of the undergrads to get familiar with ableC - what do you think, Eric?

krame505:
Ah - a couple more minor comments about `Qualifiers` after taking a closer look:
* As a convention, we use the `pps` attribute on list nonterminals instead of `pp`.  
* `getQualifiers` seems like it could just be an attribute on `Type` - any reason why not?  Also this should probably include qualifiers inside `attributedType`.

TravisCarlson:
> Duplicates should be avoided because of performance, and also printing an inferred type with a large number of repeated qualifiers (e.g. nonnull nonnull nonull int*) is suboptimal.

I don't think any of the extensions created so far can collect duplicate qualifiers. For example, nonnull is added to the qualifiers of addressOfOp; how would nonnull be added a second time to something like `&x`? In practice might not performance then suffer, if anything, by searching for duplicates whenever a new qualifier is added and never or rarely finding duplicates whose removal would improve performance? On the other hand, I strongly agree that printing of repeated inferred qualified should be avoided, but again don't see how this would happen in practice (though maybe I'm missing something).

> Doesn't len_gte suffer from the same issue as nonnull? 

The nonnull extension is a problem because it creates an error when dereferencing a pointer not qualified as nonnull. In contrast, the len_gte extension creates an error only when subscripting an array that _is_ qualified as len_gte(N). If we want the len_gte extension to strictly guarantee that all subscripts are safe (i.e. if subscripting an array not  qualified with len_gte should be an error) then this would similarly be a problem. Perhaps we would eventually like to do that, but we're even further away on len_gte than we are on nonnull. In addition to handling non-qualified code generated by extensions, we also need to handle array sizes / indices that are only known at runtime. I'm only just beginning to work on this.

> Also, the Qualifiers bit looks good, but this does need to be updated as well for the closure, string, vector, and templating extensions.

I can do this, unless one of the undergrads would like to do it as a learning experience.

> Ah - a couple more minor comments about Qualifiers after taking a closer look:

I'll try to address these soon.

ericvanwyk:
The issue with other extensions is certainly not something the new
undergrads can do.  They learned about inherited attributes on Friday.

If merging will break other extensions - such as vector or closure - then
we should not merge.

What is the status of Jenkins?  Will it not detect this problem?  Travis
can you see any trouble from Jenkins.  Can you try building the other
extensions on your qualifiers branch?

This has really got to be something we deal with sooner rather than later.

krame505:
My comments from yesterday seem not to have sent?  Anyway, here it is again:

Those commits look good.  
After further thought, in the case of nonnull, I don't think it is possible to get duplicate qualifiers, but that is only because the result of the address op wraps the parameter in a new type.  I can't think of a real example at the moment, but if you added the qualifiers to the result of something like addition, where the basic type is the same, you could end up with duplicates.  This needs a more general solution.  

Also, another issue I just realized: Even though qualifiers are being stripped in the host tree, the collection attributes for `collectedQualifiers` are still contributed, so you will end up still inferring extension qualifiers in the host tree.  We should figure out a better solution for this - in general we should avoid changing the behavior of non-forwarding productions via collection attribute as much as possible.  I will need to think about this...

krame505:
Another simplification that could be made: Move the use of `addQualifiers` from the individual op productions to `unaryOpExpr` and `binaryOpExpr`, utilizing `collectedTypeQualifiers` from these.

krame505:
After a bit of thought about the problem I mentioned previously:
The solution is probably to just make those host productions forwarding.  (Assuming you make the above change) have `unaryOpExpr` and `binaryOpExpr` become forwarding productions that forward to `qualifiedUnaryOpExpr` and `qualifiedBinaryOpExpr`, which are the same as `unaryOpExpr` and `binaryOpExpr` were except for the addition of a list of collected type qualifiers to the parameters.  The host transformation could then set these parameters to `nilQualifier()`.  
There are probably more types of expressions that we would want to have this ability than just operators (array access and function call come to mind), but that could wait until those features are actually needed.

TravisCarlson:
@krame505 is this close to what you were thinking?

By the way, what's your suggestion to deal with `substituted` occuring on Qualifier now that it is closed?

> if you added the qualifiers to the result of something like addition, where the basic type is the same, you could end up with duplicates.

The dimensionalAnalysis is closer to something like this, though still not quite. Multiplying a type qualified as `unit(m)` by another type qualified as `unit(m)` causes a `unit(m^2)` qualifier to be collected on the operator. Though `unit(m^2)` is a single qualifier, and an alternative implementation that instead generates `unit(m) unit(m) double` wouldn't fit into the framework. Maybe this real example of collecting qualifiers on a binary operator could be helpful in thinking about this.

I've been trying to think of any example of how duplicate qualifiers might be collected on addition. Consider a qualifier `hasBeenAdded` that is always collected on addition. This alone is not enough to create duplicates on something like `1 + 2 + 3` since the types on operators drop all qualifiers (though it seems we're not sure whether we should be doing this, there is a "qualifiers?" [comment](https://github.com/melt-umn/ableC/blob/develop/edu.umn.cs.melt.ableC/abstractsyntax/TypeOps.sv#L78) in `usualArithmeticConversionsOnTypes()`). So it seems the only way duplicates could be collected in this case would be if an extension collected in `addOp` more than one of: `hasBeenAdded`, the qualifiers on `lop`, and the qualifiers on `rop`. But then I don't think it seems unreasonable to expect the extension to massage the qualifiers to be collected, as dimensionalAnalysis does, and remove duplicates from the qualifiers to be collected if desired. Or I suppose more than one extension could collect the same qualifier from `lop` / `rop`, if they collected a qualifier that they did not "own" ...

Somewhat related to this is another issue that I'm going to need to handle: Is there a certain way you'd prefer flow types to be seeded?

```
aspect production mulOp
top::NumOp ::=
{
  local units :: [Pair<DimUnit Integer>] =
    collectUnits(top.lop.typerep.qualifiers ++ top.rop.typerep.qualifiers);

  -- FIXME: exceeding flow type
  top.collectedTypeQualifiers <- [unitsQualifier(units)];
}
```

krame505:
Close- the explicit host equations that strip qualifiers should be on the
non-forwarding `qualified*Expr` productions. Also these productions should
have `Qualifiers` as the parameter, not a list of `Qualifier`s.

For `substituted`, you could add a default equation `top.substituted =
top`.  For any qualifiers doing something interesting, they can override
this with an explicit propagate.  But requiring an explicit propagate for
everything wouldn't be a bad idea either- up to you.

I'm pretty sure dropping all qualifiers on operator types is a bug- so
something like the example you gave would need to be handled.

For flow types on `BinOp`s, everything may be accessed from the lop and rop
attributes.

On May 25, 2017 11:59 PM, "Travis Carlson" <notifications@github.com> wrote:

@krame505 <https://github.com/krame505> is this close to what you were
thinking?

By the way, what's your suggestion to deal with substituted occuring on
Qualifier now that it is closed?

if you added the qualifiers to the result of something like addition, where
the basic type is the same, you could end up with duplicates.

The dimensionalAnalysis is closer to something like this, though still not
quite. Multiplying a type qualified as unit(m) by another type qualified as
unit(m) causes a unit(m^2) qualifier to be collected on the operator.
Though unit(m^2) is a single qualifier, and an alternative implementation
that instead generates unit(m) unit(m) double wouldn't fit into the
framework. Maybe this real example of collecting qualifiers on a binary
operator could be helpful in thinking about this.

I've been trying to think of any example of how duplicate qualifiers might
be collected on addition. Consider a qualifier hasBeenAdded that is always
collected on addition. This alone is not enough to create duplicates on
something like 1 + 2 + 3 since the types on operators drop all qualifiers
(though it seems we're not sure whether we should be doing this, there is a
"qualifiers?" comment
<https://github.com/melt-umn/ableC/blob/develop/edu.umn.cs.melt.ableC/abstractsyntax/TypeOps.sv#L78>
in usualArithmeticConversionsOnTypes()). So it seems the only way
duplicates could be collected in this case would be if an extension
collected in addOp more than one of: hasBeenAdded, the qualifiers on lop,
and the qualifiers on rop. But then I don't think it seems unreasonable to
expect the extension to massage the qualifiers to be collected, as
dimensionalAnalysis does, and remove duplicates from the qualifiers to be
collected if desired. Or I suppose more than one extension could collect
the same qualifier from lop / rop, if they collected a qualifier that they
did not "own" ...

Somewhat related to this is another issue that I'm going to need to handle:
Is there a certain way you'd prefer flow types to be seeded?

aspect production mulOp
top::NumOp ::=
{
  local units :: [Pair<DimUnit Integer>] =
    collectUnits(top.lop.typerep.qualifiers ++ top.rop.typerep.qualifiers);

  -- FIXME: exceeding flow type
  top.collectedTypeQualifiers <- [unitsQualifier(units)];
}

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/melt-umn/ableC/pull/34#issuecomment-304191387>, or mute
the thread

TravisCarlson:
Why are you pretty sure that dropping all qualifiers on operator types is a bug? Looking at this some more, we don't drop qualifiers on pointers, which I now see could possibly create duplicates, but it's still not clear to me why dropping qualifiers on builtInTypes would be a bug.

The line I pointed out applied when both operators are builtInTypes; I'll come back to that in a minute.

Separately, a few lines down, are the cases where one operand is a pointer and the other a builtInType. Here we drop the qualifiers on the builtInType but not the pointer (there's a bug on [TypeOps.sv:L82](https://github.com/melt-umn/ableC/blob/develop/edu.umn.cs.melt.ableC/abstractsyntax/TypeOps.sv#L82), it should return `b`). Keeping the qualifiers on the pointer is the correct behavior, at least for the standard qualifiers, e.g. `const int *x = ...; int *y = x + 1` should be an error because `x + 1` should be qualified as const. Extension qualifiers that don't want this behavior could perhaps collect a qualifier to be removed from certain operators. I think this is a more interesting problem; would there ever be a case when extensions try to both add and remove the same qualifier?

So, returning to the `hasBeenAdded` example, we could collect duplicates through pointer arithmetic, e.g. `int *x = ...; hasBeenAdded int *y = x + 1 + 1;`. But if this is the worst of it, I still don't see this as being much of an issue. I'm comfortable telling users: if you want to create a qualifier that is intended to be collected on pointer arithmetic and want to avoid duplicating your qualifier in order to improve error messages and/or performance, then in your aspect productions you should check if the lop/rop which is a pointer contains your qualifier, and don't collect it if so.

Back to dropping qualifiers on builtInTypes: I don't see how this makes any difference to the standard qualifiers, so don't see how it could be a bug. The const and volatile qualifiers (ignoring `restrict` here
because it can't qualify builtInTypes) both set `appliesWithinRef=false` as I'm calling it (it seems I set this incorrectly to true on volatile, which I'll need to fix), or `level = ref` as Jeffrey Foster calls it in CQual. See Section 4.1 of _A Theory of Type Qualifiers_.

Consider `const int x = 0;`. Should `x + x` be of type `int` or `const int`? What difference would it make? Either way, `int y = x + x` should be allowed. The difference only appears when working with qualifiers setting `appliesWithinRef = true` (or `level = value`), of which none of the standard qualifiers are, so it would seem we'd be free to choose whichever behavior we prefer. Which behavior might we prefer for extension qualifiers?

Consider something similar to the type-safe enumeration extension, where an `enum_day` qualifier annotates integers. Should `FRIDAY - MONDAY` be of type `int` or `enum_day int`? It makes sense to drop the qualifiers and for `enum_day int x = FRIDAY - MONDAY;` to be an error, as the equivalent use of enum is in C++.

I can't find any mention of this issue in the papers on type qualifiers, but looking at the source of CQual, it does appear to drop qualifiers on binary operations with two builtInTypes. Not that we need to make the same design decision as others, but I do see some validation in the fact that this at least didn't jump out at the people who worked out this theory of type qualifiers as being the wrong behavior.

TravisCarlson:
Hopefully this is getting very close to being ready to merge.

> For flow types on `BinOp`s, everything may be accessed from the lop and rop attributes.

I don't understand what is meant by "everything may be accessed".

krame505:
These changes look good.  One other minor bug I noticed - `__restrict` and
`restrict` are actually the same thing, so these should be compatible with
each other.

I think your analysis of stripping qualifiers on operator types is
reasonable.  As for avoiding duplicates via pointer arithmetic, it would
probably not be too much of an issue to make the extension writer
responsible, but is there any reason why removing existing qualifiers from
the list to be added would be bad?  In fact, this could even be worked into
`addQualifiers` - this would also fix a slight bug with the closure
extension where we add the `const` qualifier to everything that is
captured, regardless of whether it is already `const`.

What I meant about setting the flow types is that the attributes `lop` and
`rop` on `BinOp` are the decorated children of the production, and so they
should be the only attributes needed for computing
`collectedTypeQualifiers`.

I do think this is pretty close to being ready to merge - thanks for being
patient!  This is a rather complex problem, and I do think you have come up
with a fairly elegant solution.

Lucas Kramer

TravisCarlson:
What are you thinking the new `addQualifiers` would look like? Here it is now:

```
function addQualifiers
Type ::= qs::[Qualifier] base::Type
{
  base.addedTypeQualifiers = qs;
  return base.withTypeQualifiers;
}
```

I'm thinking a first attempt that would at least fix the closure bug might filter out the added qualifiers that already exist in the type,  like this:

```
function addQualifiers
Type ::= qs::[Qualifier] base::Type
{
  base.addedTypeQualifiers = filter(
      \q::Qualifier -> !containsBy(qualifierCompat, q, base.qualifiers),
      qs);
  return base.withTypeQualifiers;
}
```

But this isn't enough to prevent duplicates from being added if duplicates exist in the `qs` argument. What might that look like?

The only reason I'm hesitant to change addQualifiers is to keep the code simple. Right now `addQualifiers` is a very easy-to-understand two lines.

But, if you're interested in efficiency, the above change increases the computational complexity of adding n qualifiers to a type containing m qualifiers from O(n) to O(n*m).

TravisCarlson:
> What I meant about setting the flow types is that the attributes `lop` and `rop` on `BinOp` are the decorated children of the production, and so they should be the only attributes needed for computing `collectedTypeQualifiers`.

OK, agreed. What I was trying to get at was what the preferred method of adding `lop` and `rop` to the dependencies of `collectedTypeQualifiers` is. Is creating "hackUnused" productions the way to go?

```
abstract production hackUnusedNumOp
top::NumOp ::=
{
  top.collectedTypeQualifiers := top.lop.typerep.qualifiers ++ top.rop.typerep.qualifiers;
  forwards to hackUnusedNumOp(location=top.location);
}
```

TravisCarlson:
I've pushed a "hackUnused" method of seeding the flow types. If there's a better solution, just let me know and I can revert it.

The first patch below is the solution I've been advocating to prevent duplicate consts in the closure extension. The second patch is my attempt at preventing all duplicates from within addQualifiers. I still moderately prefer the former, but don't feel that strongly about it. I can commit either one, or a third option if you'd like to suggest one.

```diff
--- a/extensions/closure/edu.umn.cs.melt.exts.ableC.closure/abstractsyntax/Lambda.sv
+++ b/extensions/closure/edu.umn.cs.melt.exts.ableC.closure/abstractsyntax/Lambda.sv
@@ -148,6 +148,11 @@ top::CaptureList ::= n::Name rest::CaptureList
               location=builtin),
             declRefExpr(n, location=builtin),
           location=builtin)));
+
+  local constVarType :: Type =
+    if containsQualifier(constQualifier(), varType)
+    then varType
+    else addQualifiers([constQualifier()], varType);
   
   top.envCopyOutTrans =
     if isGlobal then rest.envCopyOutTrans else
@@ -156,7 +161,7 @@ top::CaptureList ::= n::Name rest::CaptureList
         declStmt(
           variableDecls(
             [], nilAttribute(),
-            directTypeExpr(addQualifiers([constQualifier()], varType)),
+            directTypeExpr(constVarType,
             consDeclarator(
               declarator(
                 n,
```

```diff
--- a/edu.umn.cs.melt.ableC/abstractsyntax/TypeOps.sv
+++ b/edu.umn.cs.melt.ableC/abstractsyntax/TypeOps.sv
@@ -367,10 +367,23 @@ Boolean ::= lval::Type  rval::Type
 function addQualifiers
 Type ::= qs::[Qualifier] base::Type
 {
-  base.addedTypeQualifiers = qs;
+  base.addedTypeQualifiers = filter(
+      \q::Qualifier -> !containsBy(qualifierCompat, q, base.qualifiers),
+      uniqueBy(qualifierCompat, qs));
   return base.withTypeQualifiers;
 }
 
+function uniqueBy
+[a] ::= f::(Boolean ::= a a) lst::[a]
+{
+       return if null(lst)
+         then []
+         else if containsBy(f, head(lst), tail(lst))
+              then uniqueBy(f, tail(lst))
+              else head(lst) :: uniqueBy(f, tail(lst));
+}
+
+
 {--
  - Compute a unique identifier coresponding to the module (host or extension) that 'owns' this type
  -}
```

krame505:
For `addQualifiers`, only adding qualifiers that haven't already been
added, but assuming the parameter qualifiers are not repeated is probably
fine.  I'm not too concerned about efficiency here, since the check would
need to happen anyway by the extension writer for adding a qualifier to a
type with an arbitrary list of qualifiers, or otherwise if the type is
guaranteed to have no qualifiers, efficiency isn't an issue.
I agree that making the extension writers to this does make for simpler
code, but the thing is that these qualifier lists are actually sets, and so
it is probably more intuitive to make this duplicate-avoiding behavior
implicit.

For seeding flowtypes, that was the way to do it until recently.  Ted just
implemented explicit flowtype signatures, which are much cleaner.  The
eventual goal is to get rid of all such `hackUnused` productions.



Lucas Kramer

TravisCarlson:
How's it look? Any last comments?

Thanks,
Travis

krame505:
Just one minor comment, `uniqueBy` already exists in the standard library, called `nubBy`.  Otherwise this looks good to merge for me!

krame505:
Fixes #32.

TravisCarlson:
Great! I've replaced `uniqueBy` with `nubBy`. Last call, if no more comments I'll go ahead and merge.

krame505:
Just a minute, looks like something broke in the templating extension.

Lucas Kramer

krame505:
OK, so currently this breaks the string, vector, closure, and templating extensions.  The issue with the vector extension is related to the issues with the string and templating extensions, so don't worry about that one at the moment.  
The string extension seems to be breaking because a `const char*` can't be converted to a `char*` - this is a bug with the qualifiers type checking, since this is legal in C.  
The closure extension quit compiling fairly early on, and was only fixed recently-ish, which means I can't simply find a regression by stepping through commits.  It appears overloading is broken somehow (maybe an issue with name mangling?), but I don't have time to trace it any further.  
I have no idea what is going on with the template extension, after a bit of digging.  This quit compiling when `[Qualifier]` was replaced by `Qualifiers` everywhere, so the regression happened somewhere between that refactoring and the other extensions being updated to compile again.  

I want to help debug this, but I just don't have enough time right now since I am working an internship this summer.  I can answer questions, but I won't be able to do any debugging until next weekend or later.

TravisCarlson:
Which examples are you running that are failing? I just ran `./build` and `./make` in the string, closure, and templating extensions and they appeared ok.

> The string extension seems to be breaking because a const char* can't be converted to a char* - this is a bug with the qualifiers type checking, since this is legal in C.

Did you write that backwards? Do you mean that there is a bug such that char* can't be converted to a const char*. i.e. are you talking about `const char *a; char *b = a;` or `char *a; const char *b = a`? The former raises a warning in gcc and an error on this branch. The latter compiles without warning with both gcc and this branch. Have you pulled the latest changes?

krame505:
Running ./make in these is failing running the tests for me.  I do have the latest versions- what is being printed by ./make in string, closure, and templating?  

> are you talking about const char *a; char *b = a; or char *a; const char *b = a? 

Neither one exactly, sorry I wasn't specific enough.  I am passing a string literal into a function expecting a `char *`.  ableC is now raising an error, GCC doesn't complain.  I think that the semantics for string literals in GCC are a little more complicated than just being a `const char *`.

TravisCarlson:
After a bit of googling, it seems the type of string literals in C is `char[]` and it is only in C++ that it is const. Any reason why the solution would be more complicated than the following patch?

```diff
diff --git a/edu.umn.cs.melt.ableC/abstractsyntax/Expr.sv b/edu.umn.cs.melt.ableC/abstractsyntax/Expr.sv
index 6e13512..dc5a67e 100644
--- a/edu.umn.cs.melt.ableC/abstractsyntax/Expr.sv
+++ b/edu.umn.cs.melt.ableC/abstractsyntax/Expr.sv
@@ -61,7 +61,7 @@ top::Expr ::= l::String
   top.globalDecls := [];
   top.defs := [];
   top.freeVariables = [];
-  top.typerep = pointerType([], builtinType([constQualifier()], signedType(charType())));
+  top.typerep = pointerType([], builtinType([], signedType(charType())));
 }
 abstract production parenExpr
 top::Expr ::= e::Expr
```

Here's the output of `./make` for the string extension:

```
carl4980@coldpress:~/melt/ableC/extensions/string$ ./make 
java  -jar ableC.jar example1.xc  -I../../extensions/string/include
mv example1.pp_out.c example1.c
rm example1.gen_cpp
java  -jar ableC.jar example2-remove.xc  -I../../extensions/string/include
mv example2-remove.pp_out.c example2-remove.c
rm example2-remove.gen_cpp
cc  -I../../extensions/string/include -g -std=gnu1x -c example1.c
cc  -I../../extensions/string/include -g -std=gnu1x -c example2-remove.c
cc  example1.o   -lgc -o example1.out
cc  example2-remove.o   -lgc -o example2-remove.out
Running example1.out... 
a: abc
b: def
c: ghi
d: abcdef
d: abcdefghij
e: -1229
f: -12345
h: xyzxyzxyzxyzxyz
g: xyzxyzxyzxyzxyzxyzxyz
i: 3.14159
j: 415
k: abcd
l: "abcd"
m: "\"abcd\\n\\n\\\\\""
n: <signed int * at 0x7ffc5d87ebe4>
o: 7ffc5d87ebe4
p: <signed int * * at 0x7ffc5d87ebe8>
q: 7ffc5d87ebe8
touch example1.test

Running example2-remove.out... 
a: defgxyzxyxyxyz
b: abcdefgabcabcabcxyxy
c: abcdefgabcabcxyzabcyxyz
test: abcdefgbcxyzabcxyxyxyz
touch example2-remove.test
```

ericvanwyk:
This confusion is one reason we should push to get the jenkins branch
merged now.

Since these extensions are getting refactored soon, why not proceed with
merging qualifiers and then let the resolution of these extension work
itself out?

TravisCarlson:
Well, I'm not sure why running ./make last time didn't build with the latest changes, but I see the errors on the string and other extensions now.

Removing const from the type of string literals fixed the errors mentioned above. Now the issue seems to be related to array subscript overloading. I'll keep looking at this, but if the answer looks obvious, please comment.

```
java  -jar ableC.jar example1.xc  -I../../extensions/string/include
:-1:-1: Errors in host tree:
../../extensions/string/include/string.xh:52:18: error: expression is not an indexable type (got struct _string_s)
../../extensions/string/include/string.xh:130:30: error: expression is not an indexable type (got struct _string_s)
../../extensions/string/include/string.xh:133:30: error: expression is not an indexable type (got struct _string_s)
example1.xc:22:2: error: str of struct _string_s not defined
example1.xc:22:7: error: str of struct _string_s not defined
example1.xc:23:2: error: str of struct _string_s not defined
example1.xc:23:2: error: str of struct _string_s not defined
example1.xc:48:2: error: str of struct _string_s not defined
../../testing/build/makefile:44: recipe for target 'example1.c' failed
make: *** [example1.c] Error 4
java  -jar ableC.jar example2-remove.xc  -I../../extensions/string/include
:-1:-1: Errors in host tree:
../../extensions/string/include/string.xh:52:18: error: expression is not an indexable type (got struct _string_s)
../../extensions/string/include/string.xh:130:30: error: expression is not an indexable type (got struct _string_s)
../../extensions/string/include/string.xh:133:30: error: expression is not an indexable type (got struct _string_s)
example2-remove.xc:20:2: error: str of struct _string_s not defined
example2-remove.xc:20:2: error: str of struct _string_s not defined
../../testing/build/makefile:44: recipe for target 'example2-remove.c' failed
make: *** [example2-remove.c] Error 4
make: Target 'all' not remade because of errors.
```

TravisCarlson:
I've narrowed down the problem a bit. The string extension works if the creation of `qualifiedBinaryOpExpr` is reverted so that the original non-forwarding `binaryOpExpr` is restored. The problem seems to be related to the `binaryOpExpr` and `arraySubscriptExpr` productions in overload/Expr.sv, but I'm not familiar enough with the overloading code to solve this quickly. This is definitely something that should be fixed before merging.

krame505:
I'll take a look again tonight, if I get a chance.

Lucas Kramer

TravisCarlson:
Found a simple mistake of mine that was causing these operator overloading problems. Please confirm if there are any outstanding issues.

When replacing `propagate host` with an explicit equation to remove the collected qualifiers, I failed to access host on the other children:

```diff
diff --git a/edu.umn.cs.melt.ableC/abstractsyntax/Expr.sv b/edu.umn.cs.melt.ableC/abstractsyntax/Expr.sv
index d8d723e..1fa7c0e 100644
--- a/edu.umn.cs.melt.ableC/abstractsyntax/Expr.sv
+++ b/edu.umn.cs.melt.ableC/abstractsyntax/Expr.sv
@@ -85,7 +85,7 @@ abstract production qualifiedUnaryOpExpr
 top::Expr ::= op::UnaryOp  e::Expr  collectedTypeQualifiers::Qualifiers
 {
   propagate lifted;
-  top.host = qualifiedUnaryOpExpr(op, e, nilQualifier(), location=top.location);
+  top.host = qualifiedUnaryOpExpr(op.host, e.host, nilQualifier(), location=top.location);
   top.pp = if op.preExpr
            then parens( cat( op.pp, e.pp ) )
            else parens( cat( e.pp, op.pp ) );
@@ -269,7 +269,7 @@ abstract production qualifiedBinaryOpExpr
 top::Expr ::= lhs::Expr  op::BinOp  rhs::Expr  collectedTypeQualifiers::Qualifiers
 {
   propagate lifted;
-  top.host = qualifiedBinaryOpExpr(lhs, op, rhs, nilQualifier(), location=top.location);
+  top.host = qualifiedBinaryOpExpr(lhs.host, op.host, rhs.host, nilQualifier(), location=top.location);
   -- case op here is a potential problem, since that emits a dep on op->forward, which eventually should probably include env
   -- Find a way to do this that doesn't cause problems if an op forwards.
   top.pp = parens( ppConcat([ 
```

krame505:
Ah, that would do it. I will try to test this tonight or tomorrow.

On May 31, 2017 11:55 AM, "Travis Carlson" <notifications@github.com> wrote:

Found a simple mistake of mine that was causing these operator overloading
problems. Please confirm if there are any outstanding issues.

When replacing propagate host with an explicit equation to remove the
collected qualifiers, I failed to access host on the other children:

diff --git a/edu.umn.cs.melt.ableC/abstractsyntax/Expr.sv
b/edu.umn.cs.melt.ableC/abstractsyntax/Expr.sv
index d8d723e..1fa7c0e 100644---
a/edu.umn.cs.melt.ableC/abstractsyntax/Expr.sv+++
b/edu.umn.cs.melt.ableC/abstractsyntax/Expr.sv@@ -85,7 +85,7 @@
abstract production qualifiedUnaryOpExpr
 top::Expr ::= op::UnaryOp  e::Expr  collectedTypeQualifiers::Qualifiers
 {
   propagate lifted;-  top.host = qualifiedUnaryOpExpr(op, e,
nilQualifier(), location=top.location);+  top.host =
qualifiedUnaryOpExpr(op.host, e.host, nilQualifier(),
location=top.location);
   top.pp = if op.preExpr
            then parens( cat( op.pp, e.pp ) )
            else parens( cat( e.pp, op.pp ) );@@ -269,7 +269,7 @@
abstract production qualifiedBinaryOpExpr
 top::Expr ::= lhs::Expr  op::BinOp  rhs::Expr
collectedTypeQualifiers::Qualifiers
 {
   propagate lifted;-  top.host = qualifiedBinaryOpExpr(lhs, op, rhs,
nilQualifier(), location=top.location);+  top.host =
qualifiedBinaryOpExpr(lhs.host, op.host, rhs.host, nilQualifier(),
location=top.location);
   -- case op here is a potential problem, since that emits a dep on
op->forward, which eventually should probably include env
   -- Find a way to do this that doesn't cause problems if an op forwards.
   top.pp = parens( ppConcat([

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/melt-umn/ableC/pull/34#issuecomment-305249777>, or mute
the thread

krame505:
Looks good to me.  Go ahead and merge, if nobody else has comments.

PR #36, opened 2017-04-20
Workaround for Silver issue 139.

remexre:
This is a short-term solution. Realistically, we should:

1. Fix `imports` by fixing [Silver issue 139](https://github.com/melt-umn/silver/issues/139).
2. Rename `implode`, `concat`, etc. in `silver:langutil:pp` to `ppImplode`, `ppConcat`, etc., and name all `Document` functions as such.
3. Monoid typeclass?

### DIFF
--- a/edu.umn.cs.melt.ableC/abstractsyntax/AsmConstruct.sv
+++ b/edu.umn.cs.melt.ableC/abstractsyntax/AsmConstruct.sv
@@ -5,7 +5,7 @@ abstract production asmStatement
 a::AsmStatement ::= arg::AsmArgument
 {
   propagate host, lifted;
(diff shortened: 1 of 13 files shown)


krame505:
I don't think we should rename the functions in `silver:langutil:pp` - let users do this disambiguation however they want, or take the approach we use with treemap and request that users always import things as pp:_.  

But yes, typeclasses are the best solution, long-term.  Useful ones: Monoid as mentioned above, Monad to fix hacky do expression that is currently parameterized by bind and return functions, and an automatic typeclass corresponding to every attribute that would, for example, let us derive instances by just writing equations on a nonterminal, or write functions that work over anything with certain attributes.  

Another note - changing all the names would also require more changes to change things back once we do add typeclasses.

krame505:
Fix looks good to me.

PR #39, opened 2017-04-25
Redesign overloading mechanism to be non-interfering, better handling of GCC __attributes__

krame505:
A few things got rolled into one with this...
## Redesign operator overloading to be non-interfering
The previous mechanism for overloading suffered from interfence issues, since it required the explicit definition of attributes on forwarding types, which could be lost via transformations.  

This method of overloading works by staticly building lists of possible overloads paired with module names via collection production attributes on the dispatching productions.  Here the module name is a unique string based on the grammar name of the extension, which can be set on a typedef or struct type via GCC's `__attribute__` mechanism.  To avoid prevent unexpected behavior, we require all overloaded types to be a 'new' type not compatible with any host types, such as structs, unions, or enums.  
The module name of an operator argument, if set, can be accessed from its typerep via the moduleName attribute and used to look up the forward for a dispatching production.  For binary operators, we must look at the module names of both the left and right children, and check if there is an overload between those two extension types, an extension type on the left and a host type on the right, or an host type on the left and an extension type on the right.  

In order to define an overload, the extension writer must declare a new type with the `module(<name>)` attribute.  For example
```c
struct __attribute__((module("org:ext:foo"))) foo { ... }
```
or
```c
typedef __attribute__((module("org:ext:bar"))) union bar_u bar;
```
The extension writer also writes an aspect production for the host production that they wish to overload, and contribute a pair containing their extension module name and the overload expression to the list of overloads.  

Overloadable constructs include all numeric, logical, assignment and comparison operators, array subscript, function call, member access, assignment to array index, and call to a member access.  

### Introduce `Attributes` nonterminal to replace `[Attribute]` lists
This is a breaking change for most extensions, since there are a fair number of host productions where these are attached.  This fixes a lot of ugliness, especially where we needed to write explicit equations for functor attributes that should be propagated.

### Attach attributes on types
Attributes that are written on a typedef or should be included in the type.  This is used, for example, in specifying a module name for a new type.

This is implemented by adding a new host production `attributedType :: (Type ::= Attributes Type)`, which wraps a type with a list of attributes.  This is considered type-compatible with the wrapped type, and so any time when pattern matching on Type, `attributedType` must be handled, or the attribute `withoutAttributes` may be used to strip a type down to its non-attributed form.  

Since all types must be able to be translated back to a type expression, we must be able to generate a type expression for an attributed type.  This presents a problem, since C only allows attributed types to be written at the top level in a typedef.  Usually attributes only occur at the top level of a typedefed type so the type can be simply translated back to name, but since we allow extensions to perform arbitrary transformations on types, this isn't always the case.  We can solve this problem by using the lifting mechanism to generate typedefs corresponding to these types, with the appropriate attributes attached.  

### Modify `__attribute__` mechanism of specifying static refIds
GCC allows attributes to be attached in several places: value definitions, typedefs, type expression declarations, struct and union definitions.  Previously, we only allowed refIds to be set via GCC attribute on an actual definition of a struct.  We would like to be able to set the refId for a struct prototype, which may actually be defined separately from the actual body of the struct.  Thus we need to be able to interpret a `refId(name)` attribute attached to a typedef or type expression declaration.  
This is done by getting the refId from the attributes listed in these declarations, and passing it down BaseTypeExpr via an autocopy attribute `givenRefId`.  `tagReferenceTypeExpr`, `structTypeExpr` and `unionTypeExpr` then look at this attribute if they decide that a new refId must be created.  

This is also a breaking change, since the forward flow type for BaseTypeExpr now includes `givenRefId`.

## Implement templated struct prototypes
Now that we can specify a refId for a struct forward declaration, we can implement forward declarations of templated structs fairly easily.  We can also do away with the aweful `defsDecl` hack for setting refIds in struct template declarations, which means we can once again remove `defsDecl` via the host attribute.  Thus in the host and lifted trees, all refIds are generated.  

## Fix bug with lifting
There was a bug with lifting in that all the lifted declarations from the injected decls were put before all the injected decls, instead of each decl being directly preceded by the global decls that it emitted.  This bug was being suppressed when we removed `defsDecl` via `lifted`.  This is now fixed.  

# Things to resolve before this is merged
* Merge in ableC/silver changes to pp function names
* Finish updating other example extensions with `Attributes` instead of `[Attribute]`
* Bug: currently we lift attributed typedefs to the global level when translating attriubted types.  These should really be lifted to the closest definition scope, but we don't support this right now.  Fix?
* Attribute animation - right now we attach all unhandled to the type via `attributedType`, but there are probably some that should be not be attached, i.e. a transformed version of the type shouldn't include these attributes


ericvanwyk:
Lucas, this is good documentation.  We should be keeping this in someplace
other than Git merge messages...

krame505:
Ted okayed this, I will merge soonish unless there are other objections.  The last commit with the original overloading mechanism can be found in the old_overload tag.

Issue #40, opened 2017-05-02
Looking up an overload from outside the overloaded production is painful


krame505:
Right now we have the collection attributes for everything inside productions.  This means that we run into issues if we want to, for example, look up whether `show` is overloaded for a subtype when computing a different overload.  

This could be fixed by moving these collection attributes to overload lookup functions, and writing aspect functions instead of aspect productions to specify overloads in extensions.

krame505:
Fixed by #41.

PR #41, opened 2017-05-15
Refactor overloading to utilize aspect functions

krame505:
Collection attributes for extensions to contribute new overloads are moved to their own functions.  This allows for an extension performing error checking to explicitly call a function and see if there is an overload for a type, without needing to decorate an AST and look for errors.  The string extension was also refactored in a similar way, eliminating this awful hack that the vector extension performed to for error checking on vector `show`.  

@ericvanwyk @tedinski does this look OK?  Hopefully this will be the last major thing to change with overloading.

### DIFF
--- a/edu.umn.cs.melt.ableC/abstractsyntax/Decls.sv
+++ b/edu.umn.cs.melt.ableC/abstractsyntax/Decls.sv
@@ -479,8 +479,9 @@ top::ParameterDecl ::= storage::[StorageClass]  bty::BaseTypeExpr  mty::TypeModi
   top.errors <- name.valueRedeclarationCheckNoCompatible;
 }
 
-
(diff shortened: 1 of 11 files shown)


krame505:
TODO after talking to Eric:
* Fix bug with += style operators reusing the lvalue
* Add more comments explaining what is going on in dispatching productions
* Figure out what is going on with why overloading for string extension is working without forwarding to attributedType

krame505:
I have addressed the above notes, this also fixes #40.  

@tedinski Eric requested that you review this before merging.

krame505:
@tedinski are you able to look at this soon?  I would like to get this merged.

tedinski:
I gave it a quick glance over and it looks good. If there's something
specific you want to chat about just ask, of course.

PR #50, opened 2017-06-18
Fix flow errors related to substituion

krame505:
This branch fixes a couple of flow errors, and makes the substitution grammar an option.  We may want to make this an export instead, but I am waiting to hear from Ted his thoughts on whether option is appropriate for situations where we don't want to add things to the reference set.  

Copy from our previous email chain (in reverse chronological order):

Looking at this a bit more, it seems that the new flow errors seem to be originating from that 'substitutions' is part of the reference set when the grammar is included as an export, but not when it is included as an option.  
This does seem like a useful behavior, to be able to have 'optional' abstract syntax modules that don't add items to the reference set - the alternative is writing explicit flowtype signatures everywhere, when we may want a different default for a different 'kind' of host grammar.  Could this be a potential reason to keep the substitution grammar as an option?  

Lucas Kramer

On Sat, Jun 10, 2017 at 2:31 PM, Ted Kaminski <tedinski@cs.umn.edu> wrote:
Sounds like somewhere there's a bug in Silver where it's not following options correctly.

I'll try to look soon.

Ted

On Sat, Jun 10, 2017 at 2:29 PM, Lucas Kramer <krame505@umn.edu> wrote:
When I change this to an export, there are a whole bunch of flow errors that show up that aren't there when this is an option.  Can you take a look at this?  See Ast.sv on the fix/substitutionflow branch.  

Lucas Kramer

On Tue, Jun 6, 2017 at 8:01 PM, Lucas Kramer <krame505@umn.edu> wrote:
That is interesting- I wonder why there wasn't a flow error being raised when I didn't import the substitution grammar, if it is exactly the same as exports? Unless I am still misunderstanding. 
I will change this to be an export in the host. 

On Jun 6, 2017 6:49 PM, "Ted Kaminski" <tedinski@cs.umn.edu> wrote:
I neglected to actually answer your question about option. It's exactly like export for everything related to flow analysis. And yeah, conditional exports are meant to work with them.

But this means there's no difference here. Every extension has to deal with substituted no matter what.

Really, the only place where options are actually useful is if there's concrete syntax you want to conditionally export. For attributes and abstract productions, there's no real point in trying to leave them out in an option. The flow analysis still cares about them the same as if they were exported.

Ted

On Jun 6, 2017 6:04 PM, "Lucas Kramer" <krame505@umn.edu> wrote:
Yes, which I thought of doing, but then every extension, including simple examples, need to propagate substituted. But maybe we just need to live with that?

On Jun 6, 2017 1:47 PM, "Ted Kaminski" <tedinski@cs.umn.edu> wrote:
The problem you're trying to solve is that you don't want substituted to depend on the fwd inherited attributes, which extension attributes are required to do, right?

You could just export it from abstractsyntax, too.

Ted

On Mon, Jun 5, 2017 at 8:49 PM, Lucas Kramer <krame505@umn.edu> wrote:
I just wanted to check that my understanding of the semantics of 'option' is correct.  This has no effect on the build (other than compiling the option grammars), unless the optioned grammar is imported explicitly, correct?  
I am thinking that making it an option would be a good fit for fixing our handling of the substitution grammar.  If I understand correctly, this would mean that an extensions could extend the host without importing this grammar, and would pass the flow analysis without needing to propagate substituted on pretty much every forwarding production.  This could be nice for tutorial extensions.  
But when an extension that does import substitution attempts to compose with this extension, there will be flow errors introduced in the extension that lacks the propagate equations for substituted.  Is there any way to specify these missing equations explicitly?  Is this where conditional exports come into play?  I wasn't able to find any documentation for option anywhere, this should really be added to the reference guide.  

So I guess practically, what I am asking, is should we make the substitution grammar an option or an export from the host in ableC?  

Lucas Kramer

### DIFF
--- a/edu.umn.cs.melt.ableC/abstractsyntax/Ast.sv
+++ b/edu.umn.cs.melt.ableC/abstractsyntax/Ast.sv
@@ -5,3 +5,6 @@ imports silver:langutil:pp;
 imports edu:umn:cs:melt:ableC:abstractsyntax:env;
 imports edu:umn:cs:melt:ableC:abstractsyntax:construction;
 
+-- Treat the substitution module like an export for modular analyses,
+-- but make the user import it explicitly 
+option edu:umn:cs:melt:ableC:abstractsyntax:substitution;
--- a/edu.umn.cs.melt.ableC/abstractsyntax/Ast.sv
+++ b/edu.umn.cs.melt.ableC/abstractsyntax/Ast.sv
@@ -5,6 +5,6 @@ imports silver:langutil:pp;
 imports edu:umn:cs:melt:ableC:abstractsyntax:env;
 imports edu:umn:cs:melt:ableC:abstractsyntax:construction;
 
--- Treat the substitution module like an export for modular analyses,
--- but make the user import it explicitly 
-option edu:umn:cs:melt:ableC:abstractsyntax:substitution;
+-- This module adds a new synthesized attribute that does not depend on the forward,
+-- so it must be exported.  
+exports edu:umn:cs:melt:ableC:abstractsyntax:substitution;
--- a/edu.umn.cs.melt.ableC/abstractsyntax/Ast.sv
+++ b/edu.umn.cs.melt.ableC/abstractsyntax/Ast.sv
@@ -5,6 +5,6 @@ imports silver:langutil:pp;
 imports edu:umn:cs:melt:ableC:abstractsyntax:env;
 imports edu:umn:cs:melt:ableC:abstractsyntax:construction;
 
--- This module adds a new synthesized attribute that does not depend on the forward,
--- so it must be exported.  
-exports edu:umn:cs:melt:ableC:abstractsyntax:substitution;
+-- Treat the substitution module like an export for modular analyses,
+-- but make the user import it explicitly 
+option edu:umn:cs:melt:ableC:abstractsyntax:substitution;
--- a/edu.umn.cs.melt.ableC/abstractsyntax/substitution/Expr.sv
+++ b/edu.umn.cs.melt.ableC/abstractsyntax/substitution/Expr.sv
@@ -43,6 +43,11 @@ top::Expr ::= op::UnaryOp  e::Expr
 {
   propagate substituted;
 }
+aspect production qualifiedUnaryOpExpr
+top::Expr ::= op::UnaryOp  e::Expr  collectedTypeQualifiers::Qualifiers
+{
+  propagate substituted;
+}
 aspect production unaryExprOrTypeTraitExpr
 top::Expr ::= op::UnaryTypeOp  e::ExprOrTypeName
 {
@@ -70,6 +75,11 @@ top::Expr ::= lhs::Expr  deref::Boolean  rhs::Name
 }
 aspect production binaryOpExpr
 top::Expr ::= lhs::Expr  op::BinOp  rhs::Expr
+{
+  propagate substituted;
+}
+aspect production qualifiedBinaryOpExpr
+top::Expr ::= lhs::Expr  op::BinOp  rhs::Expr  collectedTypeQualifiers::Qualifiers
 {
   propagate substituted;
 }
--- a/edu.umn.cs.melt.ableC/abstractsyntax/substitution/Substituted.sv
+++ b/edu.umn.cs.melt.ableC/abstractsyntax/substitution/Substituted.sv
@@ -218,3 +218,25 @@ attribute substituted<NoncanonicalType> occurs on NoncanonicalType;
 attribute substituted<BuiltinType> occurs on BuiltinType;
 attribute substituted<RealType> occurs on RealType;
 attribute substituted<IntegerType> occurs on IntegerType;
+
+-- Flowtype declarations for every nonterminal
+flowtype substituted {substitutions} on
+  AsmStatement, AsmArgument, AsmClobbers, AsmOperands, AsmOperand,
+  Attributes, Attribute, Attribs, Attrib, AttribName,
+  Decls, Decl, Declarators, Declarator, FunctionDecl, Parameters, ParameterDecl,
+    StructDecl, UnionDecl, EnumDecl, StructItemList, EnumItemList,
+    StructItem, StructDeclarators, StructDeclarator, EnumItem,
+  Expr, GenericAssocs, GenericAssoc,
+  BinOp, AssignOp, BoolOp, BitOp, CompareOp, NumOp,
+  MemberDesignator,
+  NumericConstant,
+  MaybeExpr, Exprs, ExprOrTypeName,
+  UnaryOp, UnaryTypeOp,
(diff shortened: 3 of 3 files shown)


ericvanwyk:
@krame505 , just so I am clear on this.  ableC and its extensions will pass the MWDA again if on this branch (assuming they passed before)?

Although you say the use of options is "a useful behavior" I don't think it is the intended behavior - is that right @tedinski ?

If we did include the substitution as an export, is the fix just propagating the substitution attribute and decorating any extension nonterminals with some autocopy substitution inh attributes?

krame505:
Mostly. What you said is correct, but the issue with using export instead
of option is that  `substitutions` gets added to the reference set, which
causes flow errors in the host. This could also be fixed by writing a bunch
of flowtype declarations, but it seems like wanting optional exports that
don't add things to the reference set is a common pattern, so I am
wondering if this really is the correct behavior for option.
What you said about fixing the flow errors resulting from this in
extensions is correct.

On Jun 19, 2017 4:47 PM, "Eric Van Wyk" <notifications@github.com> wrote:

@krame505 <https://github.com/krame505> , just so I am clear on this. ableC
and its extensions will pass the MWDA again if on this branch (assuming
they passed before)?

Although you say the use of options is "a useful behavior" I don't think it
is the intended behavior - is that right @tedinski
<https://github.com/tedinski> ?

If we did include the substitution as an export, is the fix just
propagating the substitution attribute and decorating any extension
nonterminals with some autocopy substitution inh attributes?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/melt-umn/ableC/pull/50#issuecomment-309583128>, or mute
the thread

tedinski:
Sorry for the late reply.

The behavior we're seeing here with options and "ref sets" isn't intended, exactly, but it's also an area where we haven't yet really fully decided on what's intended.

In the thesis, I describe our heuristic for ref sets as being "all inherited attributes exported by the nonterminal's declaring grammar". The thesis lacks the concept of options. In practice, we've actually just looked up all inherited attributes that are seen to occur (i.e. the normal environment, following `imports`) which isn't the right behavior.

Normally options are supposed to behave exactly as exports for anything flow-related. It's possible we could decide to special case options for this. (i.e. default ref set heuristic is "all inherited attributes occurring on this nonterminal that are exported (but not optioned) by this nonterminal.) This might actually... be a reasonable default, since you maybe don't want to essentially _require_ an option via needing to decorate things with its attributes.

Ultimately though, this is a heuristic, not a core part of the MWDA. We could write `flowtype Expr = decorate {env};` or whatever for every nonterminal, to declare the appropriate flow types.

We still haven't really done an audit on flow types to see if they're actually getting the values we'd want them to have. (Often when writing an extension at the moment, we discover a flow problem and go fix the host language, no?) I suspect a "quality, finished" version of AbleC would have explicit flow types for a fair number of things, especially reference sets, for every nonterminal.

krame505:
Makes sense.  If options *should* be being treated the same as exports w.r.t the flow analysis, then I am curious what is causing this behavior?  

What do you suggest I do for now, since I would like to do so soon?  Merge this branch as-is, or use exports and write all the explicit flowtypes?

tedinski:
The current behavior is that the "ref sets" are determined wrong.

I'd say go ahead and merge if "attributes from options are not in the ref set" is the only behavior you're relying on. That seems like a reasonable heuristic for us to choose.

OTOH, there's a big red X here from jenkins? :p

krame505:
Yes, that is correct.

I'm not sure what is up with Jenkins... I still haven't gotten around to
setting up my account, port forwarding, and all that.  I will take a look
this weekend, but if someone wants to look at this before then that would
be fine.

Lucas Kramer

tedinski:
I glanced at it, and it might be a problem with the jenkins build scripts or the condition tables extension. @TravisCarlson can you take a look?

http://localhost:8080/job/melt-umn/job/ableC/job/fix%252Fsubstitutionflow/4/

triggered:

http://localhost:8080/job/melt-umn/job/ableC-condition-tables/job/master/61/

which failed for an odd looking reason.

```
java -jar ableC.jar simple.xc
Error: Could not find or load main class artifact.Main
```

Looks like maybe the testing script isn't looking in the right place, or something got corrupted. ( @TravisCarlson it also looks like its using `/export/scratch/melt-jenkins/custom-silver/generated` and not its local workspace to generate files? Quite possibly that's the corruption: another build deleted generated files while this one was working!)

krame505:
Hmm, yeah, shouldn't there be a generated folder for every branch?  Maybe
we could even just have each build create a generated folder with a new
name, that is then deleted when the build finishes.

Lucas Kramer

TravisCarlson:
Yes, the -G option wasn't being used in the condition-tables Jenkinsfile. Looks like it was there before but got dropped at some point. I've added it back by including it in the SVFLAGS environment variable in the Jenkinsfile and modifying the Makefiles to append the -I options to SVFLAGS.

tedinski:
New problem with Jenkins! `make[1]: 'ableC.jar' is up to date.` We should probably clean the workspace or something before building.

tedinski:
I filed https://github.com/melt-umn/silver/issues/178 to condone the behavior you're relying on here.

I'll go tweak the jenkins builds to get things to work, because @TravisCarlson is busy right now.

krame505:
Hmm... Shouldn't the Makefile catch whether anything is changed to require
a build in the first place? But then I suppose we need to test the Makefile
somehow...

On Jun 22, 2017 10:10 AM, "Ted Kaminski" <notifications@github.com> wrote:

New problem with Jenkins! make[1]: 'ableC.jar' is up to date. We should
probably clean the workspace or something before building.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/melt-umn/ableC/pull/50#issuecomment-310409804>, or mute
the thread

tedinski:
I updated the condition tables Jenkinsfile to (a) do a clean build with make and (b) clean and use a 'generated' directory within the workspace to generate files (instead of the workspace root. Makes it easier to delete everything in it.)

I haven't touched anything else (notably skeleton, because skeleton hasn't had the changes done to make it generated files in the workspace yet, so *shrug*)

krame505:
Sounds good, I will update skeleton within the next day or two.
Is this then ready to merge?

On Jun 22, 2017 3:12 PM, "Ted Kaminski" <notifications@github.com> wrote:

I updated the condition tables Jenkinsfile to (a) do a clean build with
make and (b) clean and use a 'generated' directory within the workspace to
generate files (instead of the workspace root. Makes it easier to delete
everything in it.)

I haven't touched anything else (notably skeleton, because skeleton hasn't
had the changes done to make it generated files in the workspace yet, so
*shrug*)

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/melt-umn/ableC/pull/50#issuecomment-310489379>, or mute
the thread

krame505:
I'm merging this now unless anyone has any last-minute objections.

PR #61, opened 2017-08-27
Fix/ovrld member ptr

TravisCarlson:
getMemberOverload() from under pointer if dereferencing. This makes overloading of `MyType *x; x->y;` behave similarly to `MyType x; x.y;`, for example.

There are also Jenkinsfile changes here, reverting the testing with generated/ and adding a BACK_TO_NORMAL notification.


krame505:
I suppose that this is a reasonable change.  We did originally decide explicitly not to allow any form of overloading for pointers to 'new' types, since there could often be a valid non-overloaded semantic meaning of these operators on pointers in C.  However, this is a special case because `->` is really just syntactic sugar that is equivalent to a dereference followed by a member access, so when member access is overloaded we expect the `->` operator to change as well.  

Which brings me to another point - I'm not sure if we are really handling `->` the best way possible.  If it is fundamentally an 'extension' built into the C language, then maybe we should treat it as such in the abstract syntax, and give it its own forwarding production?  This would mean that for equations (including extension analyses) written corresponding to the `->` operator, including the forward equation that is changed in this PR, we would no longer need to explicitly define the value for `->` to be the same as a dereference followed by a member access.  Also, having the `.` and `->` operators have the same production always seemed a bit awkward to me anyway, as equations for pretty much everything other than monoid attributes (e.g. `errors`, `defs`, `freeVariables`, etc. which will probably be automatic someday anyway...) have different values depending on whether there is a dereference occurring. 
@tedinski any thoughts on this?

krame505:
Btw the Jenkins tweaks lgtm.

tedinski:
> Which brings me to another point - I'm not sure if we are really handling -> the best way possible...

I think you can make what changes you think are appropriate here. The only caution I'd make it to remember to think about error messages. I'm not sure if we'd want `x->y` to actually forward to `(*x).y` because we might want error messages to be different.

I suspect the one production with a boolean is a trade-off that Clang devs (I'm guessing that's where I got it from?) made because the alternative was two productions with some duplicate code between them. (And desugaring was off the table because that wanted error messages to be accurate.)

Anyway, probably merge this?

krame505:
Hmm- yeah, we could potentially transform away a forwarding production for `->` and lose the pp in error messages... anyway, this change doesn't have to happen right now if we decide to do this at some point.  

Also, @TravisCarlson I just realized that there is a bug with this PR, before this is merged.  I'll add a comment...

edu.umn.cs.melt.ableC/abstractsyntax/overload/Expr.sv
```diff
@@ -49,11 +49,20 @@ abstract production memberExpr
 top::Expr ::= lhs::Expr  deref::Boolean  rhs::Name
 {
   top.pp = parens(ppConcat([lhs.pp, text(if deref then "->" else "."), rhs.pp]));
-  
+
+  -- get overload function from under pointer if dereferencing
+  local ty :: Type =
+    if deref
+    then case lhs.typerep.withoutAttributes of
+           pointerType(_, sub) -> sub.withoutAttributes
```
krame505: Why are you stripping attributes from `sub` here?  The attributes attached to a type are considered in module name resolution.

krame505:
Changes lgtm, but Jenkins has been failing on develop so I expect this to fail as well.  I'm good with merging this once that is sorted out.

PR #63, opened 2017-08-28
Fix/lvalues

### LINKED ISSUE #7: Missing checks for L-values in assignment
AbleC currently allows assignment to non L-values.  As a very simple example, it allows code such as
`1 = 2;`
which is obviously not valid C.  
The solution to this would involve adding a new attribute to expressions
`synthesized attribute isLValue::Boolean occurs on Expr;`
to test if something is an L-value.  This would then be checked in assignments and an error would be raised if the left hand side is not an L-value.

200sc:
This adds `isLValue` to `Expr`. In unary operations and assignment expressions, this value is checked to determine if the expression is valid. This fixes #7.

The error messages used are rough equivalents to those given by gcc, the one difference being that gcc has something like full custom errors for each unary operator when used inappropriately, whereas we just append the name of the operator. 

This doesn't yet treat arrays or const types in any special way. I believe there's existing error checks for arrays, but am not as sure about the const modifier. 

Also added are some trivial lvalue tests, of which more can be added for the mentioned arrays and const values before this gets merged. Adding these tests required bypassing the .gitignore, so it's worth considering removing the *.c condition from that. It's also worth scanning through the negative tests that already exist and pass to see if they check lvalue conditions that this should be correcting.


ericvanwyk:
Looks like some conflicts need fixing.  Regardless, do not merge today - Aug 28 - since we are working on some last minute issues for the OOPLSA paper that need  to be resolved today.

200sc:
I'd also suggest waiting to merge until the jenkins issues with the cilk extension are sorted out

krame505:
I'll try to take a look later today.

edu.umn.cs.melt.ableC/abstractsyntax/Expr.sv
```diff
@@ -94,6 +101,16 @@ top::Expr ::= op::UnaryOp  e::Expr  collectedTypeQualifiers::Qualifiers
   top.defs := e.defs;
   top.freeVariables = e.freeVariables;
   top.typerep = addQualifiers(collectedTypeQualifiers.qualifiers, op.typerep);
+  top.isLValue = case op of
```
krame505: It is better style to use attributes than pattern matches when it makes sense.  I would suggest making `isLValue` occur on `UnaryOp`, and use that here instead.

edu.umn.cs.melt.ableC/abstractsyntax/Expr.sv
```diff
@@ -94,6 +101,16 @@ top::Expr ::= op::UnaryOp  e::Expr  collectedTypeQualifiers::Qualifiers
   top.defs := e.defs;
   top.freeVariables = e.freeVariables;
   top.typerep = addQualifiers(collectedTypeQualifiers.qualifiers, op.typerep);
+  top.isLValue = case op of 
+    | dereferenceOp() -> true
+    | _ -> false
+    end;
+
+  top.errors <-
```
krame505: Minor style note - this could be expressed as one `if` expression with an and.

edu.umn.cs.melt.ableC/abstractsyntax/Expr.sv
```diff
@@ -287,6 +311,16 @@ top::Expr ::= lhs::Expr  op::BinOp  rhs::Expr  collectedTypeQualifiers::Qualifie
   
   op.lop = lhs;
   op.rop = rhs;
+
+  top.isLValue = false;
+
+  top.errors <-
```
krame505: This check really belongs in the errors for `assignOp`.

200sc: I'm not sure how I'd go about putting the error there from this production, or the check in that production.

krame505: Just add an equation for `top.errors`?  You can access `lhs` here via `top.lop`.

200sc: Ok, I didn't notice the lop and rop elements.

krame505:
With those changes everything looks good, but hold off on merging this until the tests are passing.  There also appear to be merge conflicts, so I would suggest merging develop into this branch and I'll take a look if the tests are still failing.

200sc:
The conflicts its pointing out are things I've already merged, or at least tried to. I figured it had something to do with the tests failing (like it didn't consider the merge I did as valid because the tests failed?) but I'll try again.

krame505:
The merge conflicts don't have anything to do with the tests failing -
something must have gotten messed up when you tried to merge.

Lucas Kramer

krame505:
Yeah, I don't know what happened here but that definitely isn't how you resolve a merge conflict.  Take a look at the diff in your last commit - you are supposed to edit these inserted pieces of text to make it look how you want, and then commit that.

200sc:
I just used GItHub's interactive editor, because apparently it wasn't happy with my manual merge I did on my machine.

krame505:
Looks good to me now to merge, @ericvanwyk did previously ask to hold off on this, I am assuming the OOPSLA artifact is now sorted?

tedinski:
Should be fine to commit now, yeah.

PR #64, opened 2017-08-29
Modifications to allow extension of lexer hack mechanism

krame505:
I did a bit of thinking about how we might allow the lexer hack mechanism to be used for disambiguation in extensions, and what consequences this might have with regards to composability, and it was pretty simple to implement so I went ahead and tried.  

Essentially, the only significant change to the host is making `IdentType` a closed nonterminal.  Extensions can now add their own `IdentType`s, and write their own disambiguation functions over corresponding conflicting identifier nonterminals, utilizing the same `context` environment.  

The only significant question I see here is how this relates to composability.  Having two extensions with conflicting identifier terminals could easily lead to new lexer conflicts not covered by any disambiguation functions.  However, in this case, we can simply use transparent prefixes and specifying a preference explicitly to choose one extension over the other.  

I will shortly open a pull request in the template extension that utilizes this.  @tedinski @ericvanwyk @TravisCarlson  - any comments?

### DIFF
--- a/edu.umn.cs.melt.ableC/concretesyntax/LexerHack.sv
+++ b/edu.umn.cs.melt.ableC/concretesyntax/LexerHack.sv
@@ -44,10 +44,11 @@ global typenameType :: IdentType = typenameType_c();
 disambiguate Identifier_t, TypeName_t
 {
   pluck
-
     case lookupBy(stringEq, lexeme, head(context)) of
+    | just(lh:identType_c()) -> Identifier_t
     | just(lh:typenameType_c()) -> TypeName_t
-    | _ -> Identifier_t
+    | nothing() -> Identifier_t
+    | it -> error(s"Unexpected lookup result for ${lexeme} in disambiguation function for Identifier_t, TypeName_t: ${hackUnparse(it)}")
     end;
 }
 
--- a/edu.umn.cs.melt.ableC/concretesyntax/lexerHack/LexerHack.sv
+++ b/edu.umn.cs.melt.ableC/concretesyntax/lexerHack/LexerHack.sv
@@ -2,16 +2,15 @@ grammar edu:umn:cs:melt:ableC:concretesyntax:lexerHack;
 
 imports edu:umn:cs:melt:ableC:abstractsyntax as ast;
 
-nonterminal IdentType;
+-- Nonterminal representing an enumeration of possible identifier types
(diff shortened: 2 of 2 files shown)


tedinski:
As-is, I don't like this, and I don't think it would actually work. Each extension that tried to introduce a marking-identifier-terminal would be creating an ever-increasing set of ambiguities, and I suspect the MDA might even not allow it? Alternatively, it's a hole because the MDA isn't considering it, but it will probably lead to actual parse errors.

I do think the idea of a identifier-marking-terminal might be workable, but I think it'd require explicit support from Copper. Or at least Silver (I think we could generate code at composition time), but probably still Copper too in the long run (because we want runtime parse table composition in the long run). The current notion of a disambiguation function just doesn't work for this. The extension composition process would be dynamically creating new sets of ambiguities that aren't handled.

krame505:
Right, when trying to compose multiple extensions with these marking-terminal-identifiers as-is, you would get lexer conflicts.  But these could be resolved with transparent prefixes - these require you to specify an order of preference explicitly for when the prefix isn't present, generating these missing disambiguation functions.  So the composition still happens according to our definition, although with special support from Silver/Copper we could make the composition 'better' by avoiding these conflicts that currently require transparent prefixes.  

I'm not exactly sure what this has to do with the MDA, though.  Maybe I am still not understand something.

ericvanwyk:
The end result is that one has to write a transparent prefix for each id.  In the OOPSLA paper the example contains  `template pair<int, string>`  or something like that.

What we had before, and didn't allow parsing `a<b` was `pair<int, string>`.

With this new thing we would have `template::pair<int,string>`.  So this would let one chose what to call `template`, but otherwise may be more confusing to end users.

krame505:
Yes, it is a trade-off, essentially.  Either change the syntax to always be more verbose, or let the user specify how they want to change the syntax in the occasion that multiple independent extensions want to introduce marking terminal identifiers.  In any case, from my point of view this is better than what we have now, but I do see how consistency could be considered valuable.

krame505:
After a little more thought, this actually wouldn't be an improvement, because you would also need to specify prefixes for either the host `Identifier_t` or `TypeName_t` if an explicit preference ordering is specified.  

I've been doing a bit more thinking about what Ted said - I'll go ahead and close this and open a Silver issue with some thoughts.

PR #65, opened 2017-09-04
Add tutorial section on extensible env

krame505:
See title.  Also fix a minor bug with defsDecl in 7dd1957, if we have extensions propagate lifted on their productions and we want for some reason to perform lifting without getting the host tree, this bug would have caused problems.


krame505:
I'm going to merge this if there are no objections...

ericvanwyk:
Thanks, Lucas.  This is great.  After finishing the final version of the SLE paper today I plan on turning my attention to this sort of thing.

PR #68, opened 2017-09-15
Re-enable generated builtins

krame505:
I updated the builtin generator code to compile again after the changes to `Qualifiers`, and re-created the `GeneratedBuiltins.sv` and `GeneratedIntrinsics.sv` files.  This may slow down the ableC build slightly, but it wasn't extremely noticeable.  
Also, building the initial environment is somewhat stack-intensive, so we will need to increase the default stack size when invoking ableC.  16M seems to be sufficient, this can be set by invoking the jar as
`java -Xss16M -jar ableC.jar foo.xc`
This would need to be changed in all extensions before merging this PR.  I already made this change in templating and skeleton when I was debugging, but I will hold off on updating all the other extensions until I receive feedback on this.


krame505:
Actually, 16M is probably way overkill for stack size.  The templating tests work fine at 2M, and Silver is set to use 6M by default, so I'll go with that.

krame505:
If there are no objections, I will update the failing extensions to increase the default stack size, and merge this.

krame505:
This doesn't fix all the errors related to missing builtins, btw - some things (e.g. GCC vector intrinsics such as `__builtin_shuffle`) are missing from here.  These files were pulled from Clang, right?  @tedinski any thoughts on this?

ericvanwyk:
If this doesn't have too many negative effects on things, then I am happy
to include them.

Can you say more precisely what stack size was used before and what stack
size is required now?

Thanks,
Eric

krame505:
Previously we weren't setting the java stack size at all, this is somewhat system-dependant - 1M on my machine.  The extensions I tested work with 2M, so I went with 6M to give some buffer and since that is what Silver uses.

tedinski:
Does the MWDA finish? I thought we removed these because it took over an hour.

krame505:
Yes, it only takes 15 seconds now - I think the performance fixes we made a while ago for the first version of overloading helped a lot.

krame505:
I have the extensions passing with this now, but the ableC `test` stage is failing, since this fixed some of the previously-failing tests, and some of the previously-passing ones are now failing.  Trying to take an inventory of this has been somewhat tedious due to some of the shortcomings of our test driver, such as that there is no timeout functionality, and some tests will run for 40+ minutes on my machine before failing.  I have been attempting to remedy this, but as mentioned in https://github.com/melt-umn/silver/issues/198, I can't figure out how to kill a timed-out process without killing and re-starting the whole server... any thoughts?

tedinski:
Are those long running tests caused by the increase in stack size or lack of heap space or something? I don't recall any that used to take a long time.

200sc:
I can attest that the tests take a long time for me as well-- I didn't wait them out to see exactly how long, the last time I tried, but if it was caused by a change it would have been a change older than the fix/lvalues PR.

krame505:
I have no idea why, but they've been slow for as long as I can remember, long before this change.  Those test files are just plain big, with lots of nested macro expansions...

tedinski:
Ah, are these "performance tests" for pathological cases? We could choose to just disable these for now.

krame505:
I don't *think* that these are intended to be performance tests, e.g. `tests/gcc/default/positive/struct-by-value-11_x.c`.  But do we have a way of disabling individual tests?

krame505:
For now I'm abandoning my broken attempt at refactoring the test framework to add a timeout, dumping what I have in `feature/test_timeout`.  I'm going to add a `.disabled` suffix to the test file names that are taking ridiculous amounts of time which should make the test driver ignore them.

krame505:
OK, I have it finishing in a sane amount of time now, but some of the tests are now failing that weren't previously.  I'll need to investigate before merging.

krame505:
I updated the positive test cases and the build finally passes, so I'm going to go ahead and merge this if there aren't any objections.  

This only actually fixed 2 failing tests (:-P), as it seems a bunch of the GCC intrinsics related to vector operations are missing (since we are generating these from Clang, of course... I'll open another issue), and some of the other cases that involve builtins are failing for other reasons.

tedinski:
lgtm

PR #71, opened 2017-10-05
Get jenkins green

### LINKED ISSUE #57: Jenkins fails to mkdir first time
The feature/jenkins and fix/mwda-special-specifiers branches of ableC failed to build today. The logs showed the generated/ directory did not exist even after running `sh "mkdir -p generated"`. But the builds succeeded after simply running a second time.

### LINKED ISSUE #70: Incoming flow errors
I'm in the process of finalizing the added pattern matching support for the flow analysis. I used to think this wouldn't lead to new errors, but I discovered there's a missing piece of the analysis along the way.

I tried the (currently in-progress) analysis and it so far only found one new problem:

```
Errors for edu:umn:cs:melt:ableC:abstractsyntax:substitution
 [./edu.umn.cs.melt.ableC/abstractsyntax/substitution/Attribute.sv]
Attribute.sv:60:2: warning: Synthesized equation 'substituted' exceeds flow type with dependencies on env, returnType
```

This is the relevant production:

```
aspect production appliedAttrib
top::Attrib ::= n::AttribName  e::Exprs
{
  local substitutions::Substitutions = top.substitutions;
  substitutions.nameIn =
    case n, e of
      attribName(n), consExpr(stringLiteral(s), nilExpr()) ->
        ...
    | _, _ -> ""
    end;
  top.substituted =
    ... substitutions.refIdSub ...;
}
```

Notice we're pattern matching on `e`, which means we must be able to evaluate the forwarding equation of any expressions, which presumably requires `env, returnType`.

I believe this was missed before because we're pattern matching on `Exprs` outermost which possibly has a forwarding flow type that's empty, so the outer pattern match was OK. Then matching on the inner child erroneously believed it had a proper `Decorated Expr` and assumed the ref set. But that's not a safe assumption without checking.

So with it now "looking through" pattern matching, it sees that the inner child only has `env` if the outer has `env`, so it accurately emits the dependency, which causes the flow type for `substituted` to go awry.

Anyway, I think this implementation isn't done yet, but already it's found one new flow error. Let's hope it doesn't find any more. Just a heads-up.

tedinski:
Fixes #70 
Well, sort of. It's a hack, but it's probably a hack that's just fine for now.

Fixes #57 
Apparently `checkout scm` can clean out all non-checked out files. I'm not sure if this is guaranteed behavior or... just what seems to be happening right now. So I've left in code to delete `generated` but otherwise moved its creation to a safe spot in the script. So that should be taken care of.

@krame505 let me know if the first fix is too ugly to merge. But even if it's not ideal, I'd like to merge and just have it get fixed later, if that's okay?

### DIFF
--- a/edu.umn.cs.melt.ableC/abstractsyntax/AttributeAnimation.sv
+++ b/edu.umn.cs.melt.ableC/abstractsyntax/AttributeAnimation.sv
@@ -2,6 +2,9 @@
 function animateAttributeOnType
 Type ::= attr::Attributes  ty::Type
 {
+  -- TODO: These are hacks we should probably improve upon.
+  attr.env = emptyEnv();
+  attr.returnType = nothing();
   return
     case attr of
     | consAttribute(gccAttribute(l), t) -> animateAttribOnType(l, animateAttributeOnType(t, ty))
@@ -18,6 +21,9 @@ Type ::= attr::Attributes  ty::Type
 function animateAttribOnType
 Type ::= attr::Attribs  ty::Type
 {
+  -- TODO: These are hacks we should probably improve upon.
+  attr.env = emptyEnv();
+  attr.returnType = nothing();
   return case attr of
   -- Attribs that specify new types (i.e. not type-compatible with what is wrapped)
   -- __vector_size__(num)
--- a/edu.umn.cs.melt.ableC/abstractsyntax/substitution/Attribute.sv
+++ b/edu.umn.cs.melt.ableC/abstractsyntax/substitution/Attribute.sv
@@ -50,9 +50,14 @@ top::Attrib ::= n::AttribName
 aspect production appliedAttrib
 top::Attrib ::= n::AttribName  e::Exprs
 {
+  -- TODO: These are hacks we should probably improve upon.
+  local treat_e_syntactically :: Exprs = e;
+  treat_e_syntactically.env = emptyEnv();
+  treat_e_syntactically.returnType = nothing();
+  
   local substitutions::Substitutions = top.substitutions;
   substitutions.nameIn =
-    case n, e of
+    case n, treat_e_syntactically of
       attribName(n), consExpr(stringLiteral(s), nilExpr()) ->
         if n.name == "refId" then substring(1, length(s) - 1, s) else ""
     | _, _ -> ""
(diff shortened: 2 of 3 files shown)


tedinski:
Aww, drat. Downstream failures in halide, sqlite, and vector. Looking into that now...

tedinski:
Oh, hey, looks like it's just not enough memory. I've upped the memory limits of the `silver` script a bit to help.

krame505:
I'll take a look tonight.

tedinski:
Well, that's good news: no new flow errors in any of the extensions as a result of my changes to silver. Green check mark! Yay!

edu.umn.cs.melt.ableC/abstractsyntax/substitution/Attribute.sv
```diff
@@ -50,9 +50,14 @@ top::Attrib ::= n::AttribName
 aspect production appliedAttrib
 top::Attrib ::= n::AttribName  e::Exprs
 {
+  -- TODO: These are hacks we should probably improve upon.
+  local treat_e_syntactically :: Exprs = e;
```
krame505: Why do we need a seperate local here, instead of just providing the equations on e directly?

tedinski: Equations already exist (the autocopy implicit ones) from where this production is declared.

I could be missing something about `appliedAttrib` but I'm guessing it doesn't want to always treat `e` has having an empty environment. It's just in this situation, where we're trying to extract a string constant from a particular shape subtree.

So I felt that doing things this way was less unexpected: it won't impact any other analysis where you'd expect the subtree has the normal environment available.

krame505: Makes sense - for substitution (since that is env-independent), that is probably the best we can do, anyway.

edu.umn.cs.melt.ableC/abstractsyntax/AttributeAnimation.sv
```diff
@@ -2,6 +2,9 @@
 function animateAttributeOnType
 Type ::= attr::Attributes  ty::Type
 {
+  -- TODO: These are hacks we should probably improve upon.
+  attr.env = emptyEnv();
```
krame505: This is probably OK for now, but the proper fix would be to either add `env` and `returnType` to the parameters for `animateAttributesOnType`, or change the forward set for `Attribute` and `Attrib` to be empty.  Same goes for `animateAttribsOnType`.

PR #73, opened 2017-10-06
 Use sv-nailgun for running the expected-results tests

### LINKED ISSUE #72: Use sv-nailgun for expected-results tests
I'm not sure why, but we seem to just be re-invoking the jar for each test, which is considerably slower.  Since this is the script that gets run by Jenkins, is there any reason why this is not using sv-nailgun?

krame505:
Fixes #72.  
Note that this should probably be merged after #68.


tedinski:
lgtm

PR #75, opened 2017-10-09
Fix/remove examples

200sc:
As required for the 0.1 AbleC release, this pulls all extensions out of the ableC repository itself.

At Eric's request, the extensions were moved to github.umn.edu, under ableC-old-examples. They were moved through branch filtering, so the resulting repository still maintains historical commits about the extensions.


krame505:
Seems like a reasonable approach.  Just remember, though, that github.umn.edu isn't publicly accessible, so we might want to consider that when linking from the public repo.

200sc:
We could point in the direction of github.umn instead of giving a url that won't work, and / or note that the url is private?

ericvanwyk:
I would remove the directory completely.   The commit message should indicate that these ended up on github.umn.edu somewhere.  These are all so old that I don't think people will be looking for them, but this way we've put them somewhere just in case.

200sc:
I'm planning on squashing and using that last commit message

ericvanwyk:
Sounds good to me.

Issue #69, opened 2017-09-19
Refactor/expand binaryOpExpr and unaryOpExpr


TravisCarlson:
Currently we have a single `binaryOpExpr` production in Expr.sv with children `lhs::Expr  op::BinOp  rhs::Expr`, and many productions for the binary operators in ExprBinOps.sv, which then have inherited attributes `op.lop = lhs; op.rop = rhs;`.

In the type qualifiers paper, we wanted to explain an operator overloading example as simply as possible. The feature/type_qualifiers branch replaces the use of `addOp` with an `addExpr` production that has children `lhs::Expr  rhs::Expr`. The ableC paper similarly uses a `unaryMinusDispatch` production with child `e::Expr` rather than `unaryOpExpr(negativeOp())`.

Using `binaryOpExpr` reduces the lines of code needed. Alternatively, `addExpr`/`subtractExpr`/etc. productions would need to be created in Expr.sv and overload/Expr.sv, which are longer than the `addOp`/`subOp`/etc. productions currently in ExprBinOps.sv and overload/ExprBinOps.sv. But, if this is easier to understand, might it be worth it?

I'm creating this issue to gauge opinion before deciding what to do next on the type qualifiers branch: either revert the new `addExpr`/etc. productions back to `addOp`, or convert more of the `Op` productions to `Expr` productions.

tedinski:
In general, I think we have the "op" approach for two reasons: 1. Clang did it. 2. Clang did it because there's a fair amount of shared code between each of the operators, and we don't want to copy & paste it all the time.

I know that Silver's style strongly encourages the `addExpr` approach, and things like the MWDA give us disadvantages for the `addOp` approach (references getting passed around). So I'm generally favorable to the idea of switching. But I also think we might want to seriously prioritize some kind of "traits" notion for Silver, to help deal with duplicated code, if we do go this route. It'd be nice if we could write down an "operator trait" and use that to share common implementation between all the operator expressions.

This would be a serious design problem to solve though, but it could probably be done afterwards. Duplicate code today, clean it up tomorrow.

ericvanwyk:
I think some duplication of code is acceptable if overall it simplifies the code.  Thus I support going with ``addExpr`` over the ``binaryOpExpr`` and ``addOp`` combo.

Much of the duplicated code is equations that collect errors or defs.  Some form of monoid attributes would let us write ``propagate host, lifted, errors, defs`` perhaps.

krame505:
[...]

Issue #55, opened 2017-07-27
Extensions using Jenkins library


tedinski:
Right now we have duplicated Jenkinsfiles in every extension. This makes it difficult to fix up the build & test process for each extension all at once.

Jenkins pipelines support "shared libraries" : https://jenkins.io/doc/book/pipeline/shared-libraries/

If I understand this documentation correctly, we could do this:

Create a `melt-jenkins` repo, with the organization suggested in that link (probably just using the `vars/` approach).

Then we should be able to import that library by writing:

```
@Library('github.com/melt-umn/melt-jenkins@master') _
```

Inside each extension. I suspect we should then have each extension have a `Jenkinsfile` that looks like this:

```
@Library('github.com/melt-umn/melt-jenkins@master') _

properties(ablecExtensionDefaultProperties)

buildAbleCExtension("ableC-condition-tables")
```

I expect our library would have:

* Some default values: `ablecExtensionDefaultProperties` and `silverDefaultProperties` for ableC extensions and Silver projects generally. So we don't have to repeat their descriptions and the like.
* `buildAbleCExtension` written all in one place
* `notifyBuild` doesn't need repeating everywhere
* Maybe other helpers / refactoring as needed. :) I'd like to use some of these things in refactoring the AbleC Jenkinsfile itself, and as part of writing a Jenkinsfile for Silver (finally) too.

Thoughts?

krame505:
This would be useful, for sure.  Updating these Jenkinsfiles can get ugly...
Another note if we are going to be refactoring these: Some extensions depend on others, so changes in one could cause others to fail.  We may want to have extensions that are depended on by others build some of these dependant extensions, similar to how the ableC Jenkinsfile builds other extensions as well.

ericvanwyk:
This sounds good to me.  @tedinski do you want to get started on this?

tedinski:
I'm on a trip this week, but I'll get it done next week.

tedinski:
I've posted a start. The functions can be found here:

https://github.com/melt-umn/jenkins-lib/blob/master/vars/melt.groovy

The repo otherwise just a README at the moment, basically.

I haven't started using/testing these yet. Next week. I thought I'd just invite any public comments on what I have there so far. (I'm in the middle of trying to figure out how to make `doesJobExist` actually work, without just causing security exceptions, but other than that, there's a few things there.)

[...]

PR #79, opened 2017-10-17
Feature/type qualifiers

TravisCarlson:
This pull request contains the changes discussed in Issue #69. 

- The binaryOpExpr production and BinOp nonterminal have been completely refactored into addExpr, etc.
- Binary op expressions follow a path from concrete syntax, to overloadable (renamed from overload/), to injectable (new), to host (everything previously at ableC:abstractsyntax moved here).
- Production attributes have been created to support injecting of qualifiers and insertion of runtime code.

Not addressed:
- The semantic issues raised in the thread linked above. 
- There's a lot of duplication of code here.
- The handling of operator overloading on the flattened out assign ops is no longer complete. For example, if the assignment operator wasn't overloaded then `l %= r;` would look at the overloading of the mod operator. I only implemented this behavior on a few commonly overloaded operators like addition. Lucas [had mentioned](https://github.com/melt-umn/ableC/issues/69#issuecomment-330952316) an interest in updating this after the refactoring, and probably has better ideas on how to handle this than the method here now.

These changes break every extension. I've created new branches on many of the extensions and most are now passing in Jenkins. The impact on the extensions isn't significant, but is pretty tedious.

I see Lucas has commented on the recent commits. I'll take a closer look at these later today, but still wanted to put this out now.


krame505:
I can do some work on overloading this weekend.  There is a lot of work that needs to be done there before this is merged.  

Another significant task is that since we are refactoring `BinOp`, then we should refactor `UnaryOp` and `UnaryTypeOp` in the same way, to be consistent.  This should be a bit easier than `BinOp`, however.  

I hope I don't seem like I'm re-hashing this issue too much, but I see this as very important: I still have strong reservations about including injection prior to handling overloading, at the 'syntax' level.  I don't see this as a matter of 'we're adding a new feature, but we'll figure out how to address the semantics later' - it is really more of us adding a hacky workaround for an unrelated deficiency in the host language (that is, not being able to allow extension types to act like true numeric types.)  I don't think that this deficiency will be particularly hard to address, and I don't think being able to infer qualifiers for extension types that we would like to consider 'numeric' is a critical part of our qualifier system, that we need to have immediately.  
For this reason I don't approve of merging this feature, although if we want to keep that specific addition in a seperate branch because we describe this method in the paper, that would be OK with me.  Ultimately it is up to Eric if he wants to merge this into develop anyway.   

Otherwise this generally looks pretty good, although there is a bit of cleanup that could be done in a few places.

TravisCarlson:
Here are just a couple replies for now:

>>> Instead of import .. with fooExpr as fooExprDefault, ..., we probably just want to do import ... as host.

>>> I agree. Refer to these as host:eqExpr, not eqExprDefault.

I can go ahead and do that. The reason I didn't was because I thought then everything from host/ as host: would need to be prefixed with host:. For example, signatures will need to be `top::host:Expr ::= lhs::host:Expr rhs::host:Expr`, `top.returnType` would need to become `top.host:returnType` which then wouldn't match `top.env` which doesn't need to be prefixed, etc. 

>>> Why do imports hiding all these productions that aren't used in this file, instead of just a file-local import?

The imports of host in overloadable/ is hiding productions with the same name as productions in overloadable/ExprBinOps.sv. Alternatively, overloadable/OpOverload.sv could do a file-local import, then I would think overloadable/ExprBinOps.sv would either need to do the same hiding or import host as host and add prefixes host:Expr, etc. as above. Is that prefixing preferred? Or is there another option?

krame505:
Hmm, I guess what I was thinking originally wouldn't work.  I was thinking that you could import it twice, both prefixed by `host:` and by itself hiding all those other productions with the same name.  But Silver complains when you have multiple imports of the same grammar (See https://github.com/melt-umn/silver/issues/57, I guess this never actually got fixed?)  
So I guess you can either leave it as-is or append `host:` to everything.  I would prefer not needing lots of `host:` prefixes, but since we are starting to have lots of productions with the same names in different grammars proliferating, I would recommend going with the latter.  

(Also P.S. Travis in the future could you try to respond to comments on commits in-line instead of on the main PR thread?  You can do this by clicking on the filename at the top of the comment.  Thanks!)

ericvanwyk:
Travis and I talked and we're going with prefixing lots of things with ``host``.   Seems the best of two unappealing options.

@krame505, I agree with all your concerns about this.  So you're to "re-hashing" this too much.  But I don't see many problems in moving ahead with this and fixing it in 0.2.  (To answer another questions from somewhere else, that is why we are numbering it 0.1 and not 1.0.)  We are also leaving for Vancouver on Saturday and I want to have something on the webpage that can be downloaded as "ableC 0.1".  If version 0.1.1 come quickly after that and 0.2 follows soon too - that is just fine.

So I'll approve the pull request when @TravisCarlson thinks it works - despite issues with semantic qualifiers and concerns with assignment operators.  

If it all comes back to haunt us in terrible ways then @krame505 can say "I told you so..." :)

edu.umn.cs.melt.ableC/abstractsyntax/injectable/Expr.sv
```diff
@@ -1,117 +1,116 @@
 grammar edu:umn:cs:melt:ableC:abstractsyntax:injectable;
 
-import edu:umn:cs:melt:ableC:abstractsyntax:host
-                                             with unaryOpExpr as unaryOpExprDefault,
-                                                  arraySubscriptExpr as arraySubscriptExprDefault,
-                                                  memberExpr as memberExprDefault,
-                                                  dereferenceExpr as dereferenceExprDefault,
-                                                  explicitCastExpr as explicitCastExprDefault;
+import edu:umn:cs:melt:ableC:abstractsyntax:host as host;
```
krame505: Any reason not to use a global import for the whole grammar now?

krame505:
OK, as long as this feature actually goes away soon, I suppose I can live with having it around for a short period of time.  

But if you want an ableC 0.1 release that includes this branch ready by Saturday, that might be a bit of a stretch.  A number of extensions don't currently compile on this branch (they seem to have been disabled in the Jenkinsfile?)  There is quite a bit that still must be done with overloading (which will involve breaking changes due to `AssignOp` going away, meaning more updates to every extension that uses overloading.)  There are also still a number of things in this branch that could be refactored before being merged.  I also won't be able to contribute much until after Friday.  
How critical is it to have the type qualifiers stuff released before OOPSLA?  In my opinion is better to have a more polished release with some 'coming soon' features than to try and rush the process and break things right before we present.

edu.umn.cs.melt.ableC/abstractsyntax/injectable/Expr.sv
TravisCarlson: I guess not. I was thinking that injectable/Mods.sv importing host without the prefix might cause a problem due to it being imported twice, but it looks like not since one import qualifies it. I'll change that.

krame505: Actually that is an issue, as there is currently a bug with Silver related to duplicate imports, which we decided (I think?) to eventually disallow: https://github.com/melt-umn/silver/issues/57
I suppose that is a valid reason to use `import`, or you could change it in this file as well to be consistent.

krame505:
I don't remember if I mentioned this at some point, so apologies if this is duplicated.  
It is less than readable to have a large-ish case statement nested inside multiple function calls in the operator overloading forwards.  It would be good to move these case statements to seperate locals (maybe called `fwrd` or `resolved` or something), and just have the forwards to be a consistent
```forwards to host:wrapWarnExpr(lerrors, host:wrapQualifiedExpr(injectedQualifiers, transform, top.location), top.location);```
Also there is no need to have `option1` separated out into a separate case expression - we should just be able to match on `lhs` directly in the main case, which would make following the flow of control here slightly easier.  

@TravisCarlson if you don't have time to go back and change these right now, then I will try to get around to this in the next few days.  But just something to keep in mind when refactoring `UnaryOp` overloading.

**Comment by krame505** [comment]
Another aside - the type `(Expr ::= Expr Expr Location)` seems to be showing up in a lot of signatures.  We may consider type aliasing `AssignOp` to this in order to make things more readable (and document what this type does as well.)

**Comment by TravisCarlson** [comment]
OK, all of the extensions tested in the Jenkinsfile on the develop branch are now being testing on the type_qualifiers branch and passing. I agree there are still issues, and many opportunities for improvement, but hopefully this brings us closer.

**Comment by krame505** [comment]
Hmm, I guess merging this will require changing the Jenkinsfile back?  The best approach when we are ready to merge is probably to merge all the other extensions first, then revert the Jenkinsfile here to build develop on all the extensions, then merge this?  
Would it be possible to change the ableC Jenkinsfile to try and check out a branch with the same name as the current ableC branch when building the extensions, and if it doesn't exist then fall back to building the develop branch of the extensions?

**Comment by TravisCarlson** [comment]
Yes, we'll need to change the Jenkinsfiles back, both in ableC and in all of the extensions. Everything will fail one time, but we'll be expecting it, and will fix it quickly. Updating the Jenkinsfile to automatically checkout branches of the same name would be great, but I wouldn't have the time to look into now.

Jenkins isn't passing now, but these really don't look like real failures with ableC or the extensions. The failing extensions were able to pass with no change after retrying.

(This should probably be moved to another thread:)
Jenkins is complaining about /var/lib/jenkins/ on coldpress being almost full -- it certainly is -- and is warning that strange failures may occur -- which seems to be happening. But I don't have permission to remove the old workspace files and don't see any option in Jenkins to remove them.

```
Filesystem      Size  Used Avail Use% Mounted on
/dev/sda1        74G   69G  826M  99% /
```

ericvanwyk:
@tedinski, can you remove some of these files in /dev/sda1 or should we ask systems to do?

tedinski:
This problem should be temporarily solved. Longer term solution would be to do some workspace clean up as part of the build process for extensions. That can wait until we figure out that jenkins library thing in  https://github.com/melt-umn/ableC/issues/55

TravisCarlson:
OK, the tests are passing again. Thanks Ted.

After discussing with @ericvanwyk and @tedinski , I plan to merge this soon, leaving @krame505 's valid concerns outstanding for now, which are (let me know if I'm missing any):

- reservations about including injection prior to handling overloading, at the 'syntax' level
- UnaryOp should be refactored in the same way as BinaryOp
- overloading of gte should infer an overload from lt if it exists, same for lte/gt
- functions in ExprHelpers.sv should be marked as deprecated
- improve readability of case statements: move case statements out of forwards to separate locals, combine expressions in option1/forward
- type alias `AssignOp` to `(Expr ::= Expr Expr Location)`

krame505:
I think that list is mostly complete, although I really think at least the `UnaryOp` refactoring should happen before this is merged.  Anyway, I'll probably spend a while working on this Saturday... If this does get merged before then I will open another PR.

krame505:
Oh, one more TODO thing with this branch... the tutorials need to get updated.

PR #82, opened 2017-10-22
Clean up stuff related to type qualifiers

krame505:
This has most of the further changes/cleanup discussed in #79.  Not quite ready to merge yet, but I think this is feature-complete so I'm opening now.  

Stuff done in this branch:
* UnaryOp refactored in the same way as BinaryOp.
* Fix default overloads of lte/gte/gt, there were also some older bugs present here.
* Mark functions in ExprHelpers.sv as deprecated.
* Restructure overloadable productions to move all repeated overloading logic to seperate functions.
  * Also fixes a bug where we were overloading based on the types of `lhs` and `rhs` instead of their types after being transformed - we need to do 2 decorations here, unfortunately, as a side-effect of doing injection prior to overloading.  
* Introduce `UnaryProd :: (Expr ::= Expr Location)` and `BinaryProd :: (Expr ::= Expr Expr Location)` type aliases to make production signatures more readable.
* Other random cleanup.

Stuff yet to do before this is merged:
* Update tutorials.
* Update other extensions so they compile on this branch (@TravisCarlson I would appreciate help with this if you have time.)  

Stuff happening in a future PR:
* Get rid of injection on overloadable productions.

PR #84, opened 2017-10-23
Automatically build the correct extension branches

krame505:
This now attempts to build an extension branch with the same name as the ableC branch, and if that fails it falls back to building whatever default is specified.  
Note that this was branched off of feature/type_qualifiers, so I am pulling against that branch instead of develop, and will merge before #82 is merged.  
I'm going to go ahead and merge this once Jenkins passes.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -68,7 +68,7 @@ node {
     stage ("Extensions") {
       parallel(
         "ableC-skeleton": {
-          echo scm.branches[0].name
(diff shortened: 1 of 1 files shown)


krame505:
Oh, lol.  This is going to fail because the branch name isn't 'feature/type_qualifiers', so Jenkins is building develop for all the extensions since it doesn't match.  I'm going to just go ahead and merge anyway, since it should work in the feature/type_qualifiers branch.

PR #85, opened 2017-10-25
AbleC 0.1.1 release

krame505:


krame505:
I think the only thing we need to do here is add a HISTORY file, if we decide to do so.  
However I'm not 100% convinced that we actually need one, since github tracks release history and accompanying notes for us.  
@ericvanwyk thoughts?

krame505:
I went ahead and renamed the tags in Silver as we discussed to `v<version number>`.  We seem to have already been following that convention in Copper, so yay we're being consistent now.

tedinski:
Imo, if you think there is a better way, then History is unnecessary. Just
as long as there is some resource mentioning important changes / release
notes.

krame505:
Unless @ericvanwyk has any objections I'll go ahead and do it that way, then... FWIW Copper does the same thing, and doesn't have a HISTORY file.

ericvanwyk:
I'm not that committed to a HISTORY file - but it seems like a nice thing to have.  Can we generate it from all this GitHub info?

krame505:
I guess we could link to the github releases page from the README?  I guess I don't see a huge advantage to specifically having it in the tarball.  If you are concerned about people knowing what version they downloaded, the name of the tarball and contained folder contains the version already.

PR #91, opened 2018-03-15
Renaming of Init.

remexre:
Renames `Init` to `InitItem` and `init` to `initItem`. See https://github.com/melt-umn/silver/pull/221.

### DIFF
--- a/edu.umn.cs.melt.ableC/abstractsyntax/construction/FoldHelpers.sv
+++ b/edu.umn.cs.melt.ableC/abstractsyntax/construction/FoldHelpers.sv
@@ -26,7 +26,7 @@ GlobalDecls ::= l::[Decl]
 }
 
 function foldInit
-InitList ::= l::[Init]
+InitList ::= l::[InitItem]
 {
   return foldr(consInit, nilInit(), l);
 }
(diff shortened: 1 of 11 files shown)


krame505:
I'll take a look tonight

remexre:
Yup, I still have to fix+PR every extension that uses `init` too, so no rush.

tedinski:
If you create branches of the same name, jenkins should spot them to build along with this branch.

You probably don't need to pull request them separately. If this branch builds successfully, then we can just mass merge them.

remexre:
Looks like SQLite fails with a stack overflow: http://coldpress.cs.umn.edu:8080/job/melt-umn/job/ableC-sqlite/job/init/2/console

tedinski:
I don't really know what to make of that. It should be getting `-Xss8M` from the `silver` script, so if it's exhausting that much and not infinite looping, yiiikes. We'd probably have to figure out how to make the flow analysis take less stack space. Might not be an easy fix. Unless we just bump up stack space even more for now...

krame505:
Just took a look, we should probably also rename `InitList` -> `InitItemList`, `designatedInit` -> `designatedInitItem`, and `foldInit` -> `foldInitItem` to stay consistent.  Looking at this now, I guess I would have prefered to just rename the `init` production to `positionalInit` (leaving everything else alone) as Ted seems to have prefered, if that isn't too big of an inconvenience at this point... oh well.  

Also just a side note for future reference, I noticed (as I think you said before) that your editor is deleting trailing whitespace automatically.  This isn't a huge deal (although having the whitespace on blank lines helps with auto-indent in emacs and Eclipse), but the real problem is that this makes PR diffs really hard to see what actually changed - thanks.

remexre:
Closed in favor of https://github.com/melt-umn/ableC/pull/94.

PR #93, opened 2018-03-16
Fix bug with line numbers when whitespace follows cpp tags

### LINKED ISSUE #51: Error-reporting of line numbers off by number of comments lines?
To reproduce:

```
git clone git@github.umn.edu:melt/ableC-tensors.git
cd ableC-tensors
git checkout 4050ac2
cd test
make
java -jar ../examples/ableC.jar test.xc -I ../include
```

The error is on line 32 (using a function-call-like concrete production `ones` in a declaration), but ableC reports it as being on line 25.

krame505:
Fixes #51, finally.

The issue was that the cpp tag regex ends with matching an arbitrary number of newlines, which ends up matching the blank lines immediately following a header include - which are then ignored because the line number is reset *after* these are seen by copper.

### DIFF
--- a/edu.umn.cs.melt.ableC/concretesyntax/cppTags/CPPTags.sv
+++ b/edu.umn.cs.melt.ableC/concretesyntax/cppTags/CPPTags.sv
@@ -28,7 +28,7 @@ terminal Hash_t '#';
 
 
 -- I give up!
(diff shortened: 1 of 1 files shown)


tedinski:
If you're going to use `?` then `\r` comes before `\n`. Otherwise, looks good.

krame505:
Whoops, fixed.

PR #96, opened 2018-03-28
Fix global decl lifting bug

krame505:
Very small but important fix: Previously we were only inserting lifted declarations at the level of `GlobalDecls`.  This has a slight problem: If an extension forwards to, e.g `decls(consDecl(a, consDecl(b, nilDecl()))`, and `b` wants to lift something, it gets inserted before `a` which is a problem if the thing being lifted references something defined by `a`.  This was a fun one to track down...

Instead, we can reuse the `unfoldedGlobalDecls` attribute that translates a `Decls` nonterminal into a list of `Decorated Decl`s, including their respective `globalDecls` before each.


krame505:
I'm merging this now.

PR #99, opened 2018-04-01
Refactoring of grammars locations

### LINKED ISSUE #98: Move source grammar to a `grammars` directory
This would make things consistent with just about every other major Silver project, AFAIK.  Changing this will be annoying and break things, but waiting will make the problem worse... 

Also, I soon will want to be introducing a new grammar for the silver-ableC extension, which probably belongs under `edu.umn.cs.melt.exts.silver.ableC` and not part of the host language grammars - it would be good not to have these both sitting at the top level of the repo.  IDK, maybe the `builtins` processor grammar could move there, too.  

This is something I want to fix soon-ish, before diving into the implementation of the silver-ableC extension... so, thoughts?

krame505:
Fixes #98.  

There were a few misc. scripts that needed updating as well.  Also some slight refactoring/renaming that I've been meaning to do for a while is happening in the extensions, take a look at ableC-lib-skeleton for example.  

Also I moved the builtins processor code to `edu:umn:cs:melt:ableC:tools:builtins`.  I noticed that this code actually broke when Travis added location to `Qualifier`s since `hackUnparse` doesn't include these in its output.  It just so happens that we now have a nice reflection library to address this problem the right way!  So I went ahead and fixed that as well.  

TODO before merging: finish updating all the extensions.


krame505:
Hmm, so it looks like a change I made in `jenkins-lib` to fix fast builds on this branch broke them on develop for extensions that don't have a `tests` directory... so basically we should merge this ASAP.  https://github.com/melt-umn/jenkins-lib/issues/8 means that Jenkins will keep failing on this branch (unless we rename `ableC_sample_projects` to `ableC-sample-projects` - I vote we do this now), but it should succeed on develop.  

Also, I finished updating all the extensions in respective feature/rename branches... so, thoughts/comments/whatever before I merge this?

ericvanwyk:
No - if merging fixes things that is OK by me.  But is there a bug in the jenkin-lib to fix?

krame505:
2 different issues in play here, actually:
* Bug where we don't trigger downstream builds on the correct branch for repos with underscores in the name (https://github.com/melt-umn/jenkins-lib/issues/8).  This branch should build once I rename that repo, but we really should still fix this bug.  
* I renamed the `test` folder in all the extensions to `tests` for consistency with ableC and Silver, forgetting that this would cause problems with Jenkins fast builds on `feature/rename` where we copy the `examples` jar to the `test` folder.  I fixed this by changing the destination folder name to `tests` in `jenkins-lib`, but then this made all the extensions later fail in `develop` for the same reason.  This will be resolved once I merge this branch into `develop` so Jenkins sees the `tests` folders.

ericvanwyk:
OK - sounds good.  Thanks for the clarification.  Do what you need to do.

krame505:
Got this branch building on Jenkins finally, merging now.

PR #100, opened 2018-04-05
Add env items in function decls for standard-ish builtin identifiers

krame505:
For now, just the `const char *` values  `__func__`, `__FUNCTION__`, and `__PRETTY_FUNCTION__`, all available within functions in GCC.  Needed in order to get programs using `assert()` from `<assert.h>` to compile.


krame505:
I'm merging this now, nothing here is too controversial.

PR #103, opened 2018-04-16
Some fixes related to arrays

krame505:
I fixed a few issues related to arrays that I found while adding support for to capturing variable-length arrays in the closure extension: 
* Implement proper typing for `arrayTypeExprWithExpr`.  Previously, we were treating all `arrayTypeExprWithExpr`s as having `variableArrayType`.  I implemented `integerConstantValue` on `Expr`, to detect when an expression is just an integer constant, properly handling the case where an array has `constantArrayType`.  This currently does discard suffixes and breaks if the given constant is larger than a Silver `Integer`.  Also `char` constants are also technically allowed to be array sizes... but I'm not implementing that right now.  
* Dynamic array sizes mean that translating a `Type` can yield something with free variables, unfortunately.  When a var is referenced, also include any free variables in its `Type`.  
* Properly handle parentheses for array and function pointers.  When we see one of these in the wild, we have been correctly parsing it as `pointerType(parenType(arrayType(...)))`.  But an extension can directly forward to `pointerType(arrayType(...))`, which we need to notice and add the intervening parens explicitly.

### DIFF
--- a/grammars/edu.umn.cs.melt.ableC/abstractsyntax/host/Expr.sv
+++ b/grammars/edu.umn.cs.melt.ableC/abstractsyntax/host/Expr.sv
@@ -70,7 +70,7 @@ top::Expr ::= id::Name
   top.globalDecls := [];
   top.defs := [];
   top.typerep = id.valueItem.typerep;
-  top.freeVariables = [id];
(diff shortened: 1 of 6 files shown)


krame505:
Merging this now, since nothing here is too controversial.

PR #104, opened 2018-04-18
Add a way of accessing all but the global env scope

krame505:
Needed to fix a bug in the closure extension.

### DIFF
--- a/grammars/edu.umn.cs.melt.ableC/abstractsyntax/host/Stmt.sv
+++ b/grammars/edu.umn.cs.melt.ableC/abstractsyntax/host/Stmt.sv
@@ -9,7 +9,7 @@ abstract production nullStmt
 top::Stmt ::=
 {
   propagate host, lifted;
-  top.pp = notext();
+  top.pp = semi();
   top.errors := [];
   top.globalDecls := [];
   top.defs := [];
--- a/grammars/edu.umn.cs.melt.ableC/abstractsyntax/env/Env.sv
+++ b/grammars/edu.umn.cs.melt.ableC/abstractsyntax/env/Env.sv
@@ -83,6 +83,11 @@ Decorated Env ::= e::Decorated Env
 {
   return decorate globalEnv_i(e) with {};
 }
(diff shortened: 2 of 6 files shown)


grammars/edu.umn.cs.melt.ableC/abstractsyntax/host/Stmt.sv
```diff
@@ -9,7 +9,7 @@ abstract production nullStmt
 top::Stmt ::=
 {
   propagate host, lifted;
-  top.pp = notext();
+  top.pp = semi();
```
tedinski: What circumstances? Can those be fixed instead? I think we should at least try for our pretty printing to be readable, and wouldn't this introduce a lot of extraneous semicolons?

krame505: The issue I found is `label: ;`, which translates to `labelStmt(..., nullStmt())` and `pp`s back to just `label:`, which is a syntax error if it occurs at the end of a block.  Handling this any better would either require a new analysis that checks whether something is a `nullStmt()` without decoration, or adding a new `Stmt` production seperate from `nullStmt()` that is just a semicolon - which would be a lot of work and means extensions could forward to something without errors that `pp`s to a syntax error.  
Anyway, we already take a similar approach of being conservative about `pp` such that no AST can yield a syntax error - such as how we wrap the body of every function in an extra set of braces, or wrap everything in parentheses... anyway let me know if you have a better idea.

krame505: @tedinski Unless you have any other ideas, I would like to merge this soonish.

krame505: @tedinski I'm going to merge this today unless you object.

tedinski: Just asking about the approach, not objecting, go for it.

PR #108, opened 2018-05-10
Use new features of jenkins-lib

tedinski:
Assuming this builds successfully, this should just be an update to the Jenkinsfile style, like https://github.com/melt-umn/silver/pull/249

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -4,10 +4,8 @@ library "github.com/melt-umn/jenkins-lib"
 
 melt.setProperties(silverBase: true)
 
-node {
-try {
-
-  def ABLEC_BASE = pwd()
+melt.trynode('ableC') {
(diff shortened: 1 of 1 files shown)


krame505:
@tedinski I just noticed that the downstream integration builds for this are very broken since this change, as it is simply building one extension a whole bunch of times.  I have no idea why....

tedinski:
Oh my. Groovy is an even more terrible language than I thought. It's capturing a mutable loop variable by reference.

Do you have time to fix it? You can just convert it back to function calls to a `task_whatever` function that returns a closure (the braces), instead of directly constructing the closure in the loop. Otherwise, I can get to it thursday.

Annoying. Silver's jenkinsfile probably needs fixing too.

krame505:
I took a stab at fixing this and SIlver, let me know what you think.

Lucas Kramer

PR #110, opened 2018-06-02
Fix Jenkins downstream integration builds

krame505:
Apparently groovy captures loop variables by reference.  Instead, the standard pattern to use when constructing a list or map is to use `collect` or `collectEntries`.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -20,7 +20,7 @@ melt.trynode('ableC') {
     }
   }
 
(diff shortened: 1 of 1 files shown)


krame505:
Looks like something broke when I merged this - I can't access Jenkins now, but I can take a look tonight.

tedinski:
I spotted an unrenamed string in ablec-sample-projects' Jenkins file. I corrected it.

Issue #113, opened 2018-08-13
C Preprocessor Invocation


counc009:
In the ableC compile driver, the C preprocessor is invoked with the -D _POSIX_C_SOURCE argument. Without providing a value to the -D option, this automatically sets _POSIX_C_SOURCE to the value of 1, resulting in POSIX functionality from 1990 (POSIX.1, IEEE Standard 1003.1-1990) being made available. This seems like a very limiting option, as we have found a few features that are not supported in this standard, including read-write locks in pthreads, and some timing features.
We were wondering whether there is a reason that the flag is set the way it is, instead of using a more recent POSIX standard? 
Thanks.

krame505:
It appears that I am to blame for this, although I don't exactly recall why it was needed.  IIRC there was a library we wanted to use for an extension that was trying to do something buggy/unsupported in ableC and I found this as a workaround to get things to compile temporarily.  If you figure out how to make it work and all the downstream extensions still build OK, then feel free to change this.

counc009:
I've submitted a pull request, the fix I used is passing Jenkins and resolved the issues we were having.

PR #114, opened 2018-08-15
Update preprocessor invocation to POSIX 2009

counc009:
As described in issue 113 (https://github.com/melt-umn/ableC/issues/113), we've run into a few issues with the POSIX version used by the C preprocessor. This updates the POSIX version to 2009, which seems to resolve our issues.


krame505:
Everything seems to be working, go ahead and merge!

PR #115, opened 2018-08-15
Update Expr.sv to allow integration of '<-' at '==' precedence

comme043:


comme043:
I just looked closer and realized the Channels extension needs support for the assignment operator precedence level, not "==", so disregard this unless the change would be helpful in general.

PR #116, opened 2018-08-15
Update Expr.sv to include unary op at assignment precedence

comme043:


TravisCarlson:
Lucas,
I asked Ciaradh to look into adding a nonterminal for an operator at the same precedence as assignment for use with the Go-style Channels extension which introduces `<-` as both a binary operator and unary prefix operator. I wasn't clear which one I was asking about, didn't know that  `AssignOp_c` already existed (I assumed the Go extension didn't use it because it didn't exist), and so Ciaradh must have seen that and assumed I was asking about the unary operator rather than the binary one. It looks like ableC has everything we should need for the Channels extension.

Ciaradh,
Thanks for looking into this. If you still have time I could still use your help on the Channels side looking into whether the `<-` concrete syntax can be improved.

krame505:
OK, sounds good.  I'm going to delete this branch unless there are any objections.

PR #120, opened 2018-09-27
Feature/type qualifiers

TravisCarlson:
This makes function calls injectable for use with the watch type qualifier as illustrated in the COMLAN paper.


grammars/edu.umn.cs.melt.ableC/abstractsyntax/injectable/Mods.sv
```diff
@@ -109,21 +109,22 @@ top::RuntimeMods ::=
 function applyMods
 Expr ::= l::[RuntimeMod] e::Decorated Expr
 {
+  -- copy to tmp to prevent being evaluated more than once, but only if type is known (it is not anon)
+  local createTmp :: Boolean = true;
```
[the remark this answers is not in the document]
TravisCarlson: Given an expr `e` of type `union foo *`, for example, this function creates a declaration `union foo *tmp = e;` and uses `tmp` in place of `e` to avoid evaluating `e` more than once. But if `e` is instead a pointer to an anonymous union/struct then the generated declaration is the erroneous `union <anon> *tmp = e;`. I intended to skip the `tmp` declaration if the type is an anonymous union/struct, but didn't see anything on `e.typerep` to indicate this. I ended up working around this by naming the unions in the examples that used this.

grammars/edu.umn.cs.melt.ableC/abstractsyntax/injectable/Expr.sv
```diff
@@ -65,3 +75,99 @@ top::host:Expr ::= ty::host:TypeName  e::host:Expr
         top.location),
       top.location);
 }
+
```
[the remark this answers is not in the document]
TravisCarlson: Could you please take a look at the updated code?

grammars/edu.umn.cs.melt.ableC/abstractsyntax/injectable/Mods.sv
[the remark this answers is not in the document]
krame505: This is a known issue with the way we translate anonymous structs and unions in general - an extension can forward to `directTypeExpr(some anon struct type)` and invalid code gets pp'ed.  The more general fix is probably to somehow rewrite all anon structs and unions to have a generated name, but we'd need to give this some thought and it isn't super critical.  So for now I would avoid trying to handle it specially here.

grammars/edu.umn.cs.melt.ableC/abstractsyntax/injectable/Expr.sv
```diff
@@ -37,10 +37,20 @@ top::host:Expr ::= lhs::host:Expr  deref::Boolean  rhs::host:Name
   production attribute injectedQualifiers :: [host:Qualifier] with ++;
   injectedQualifiers := case top.env, top.host:returnType of emptyEnv_i(), nothing() -> [] | _, _ -> [] end;
 
+  -- to ensure modified lhs is an lvalue in case this memberExpr is enclosed in
```
krame505: Missed this on my first pass - I don't quite understand the purpose of this?  If `x.m` isn't an lvalue then `x` isn't an lvalue, and `&x` would be an error.

grammars/edu.umn.cs.melt.ableC/abstractsyntax/injectable/Expr.sv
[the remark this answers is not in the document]
krame505: Looks good.

grammars/edu.umn.cs.melt.ableC/concretesyntax/Declarations.sv
```diff
@@ -317,14 +317,11 @@ concrete productions top::DirectDeclarator_c
...
 
--- TODO: any better proposal for concrete syntax for qualifiers on functional types?
-terminal FunctionQualifiers_t '__function_qualifiers' lexer classes {Ckeyword};
-
 closed nonterminal OptTypeQualifierList_c with location, typeQualifiers;
 concrete productions top::OptTypeQualifierList_c
 |
     { top.typeQualifiers = ast:nilQualifier(); }
-| '__function_qualifiers' '(' q::TypeQualifierList_c ')'
+| '<' q::TypeQualifierList_c '>'
```
krame505: I don't really like this syntax, especially as a part of the host - it makes it seem like these are type parameters somehow.  Any reason not to go with the C++ syntax, and just allow a bare list of qualifiers after the function parameters?

TravisCarlson: I'd prefer a bare list of qualifiers but had trouble solving ambiguities with the old K&R syntax for function declarations as below. I'm also open to other suggestions.

```
void foo(x, y) int x; char *y; {
  ...
}
```

krame505: Ah, I forgot about that, annoying.  But presumably we don't care about specifying qualifiers on K&R-style functions.  We could tweak the grammar a bit:
* The conflict doesn't show up to have a bare `TypeQualifierList_c` on `DirectAbstractDeclarator_c`
* For function pointer typedefs/declarations, we can do the same thing with `DirectDeclarator_c` if we no longer use `Declarator_c` on function declarations.  Possibly duplicate `Declarator_c` and `DirectDeclarator_c` as `FunctionDeclarator_c` and `DirectFunctionDeclarator_c`, versions that don't allow a trailing `TypeQualifierList_c`
  * Alternatively, if you don't want to complicate the grammar, you could just require the `TypeQualifiersList_c` to be wrapped in parentheses or something on `DirectDeclarator_c` - this would only show up in function pointer typedefs/declarations, in practice.
* Add new productions on `InitialFunctionDefinition_c` that have a `TypeQualifierList_c` instead of a `DeclarationList_c`

grammars/edu.umn.cs.melt.ableC/abstractsyntax/injectable/Expr.sv
TravisCarlson: The issue is if the _modified_ `x.m` containing the injected code isn't an lvalue. The purpose is to replace `({T tmp = x; ...; tmp;}).m = rhs;` with `({T *tmp = &x; ...; tmp})->m = rhs;`. As in the following example:

```c
struct Foo {
        int m;
};

int main(void) {
        struct Foo x;

        // error: lvalue required as left operand of assignment
        ({ struct Foo tmp = x;
           tmp;
        }).m = 6;

        // OK
        ({ struct Foo *tmp = &x;
           tmp;
        })->m = 6;
}
```

krame505: Ah, OK.  Makes sense.

grammars/edu.umn.cs.melt.ableC/concretesyntax/Declarations.sv
TravisCarlson: Is this at all close to what you were thinking?

```diff
diff --git a/grammars/edu.umn.cs.melt.ableC/concretesyntax/Declarations.sv b/grammars/edu.umn.cs.melt.ableC/concretesyntax/Declarations.sv
index 06b9f52..5ac0dea 100644
--- a/grammars/edu.umn.cs.melt.ableC/concretesyntax/Declarations.sv
+++ b/grammars/edu.umn.cs.melt.ableC/concretesyntax/Declarations.sv
@@ -89,6 +89,20 @@ concrete productions top::Declarator_c
       dd.givenType = top.givenType;
       top.ast = dd.ast; }
 
+closed nonterminal FunctionDeclarator_c with location, declaredIdent, declaredParamIdents, ast<ast:TypeModifierExpr>, givenType; 
+concrete productions top::FunctionDeclarator_c
+| p::Pointer_c dd::DirectFunctionDeclarator_c
+    { top.declaredIdent = dd.declaredIdent; 
+      top.declaredParamIdents = dd.declaredParamIdents;
+      p.givenType = top.givenType;
+      dd.givenType = p.ast;
+      top.ast = dd.ast;}
+| dd::DirectFunctionDeclarator_c
+    { top.declaredIdent = dd.declaredIdent;
+      top.declaredParamIdents = dd.declaredParamIdents;
+      dd.givenType = top.givenType;
+      top.ast = dd.ast; }
+
 
 closed nonterminal TypeName_c with location, ast<ast:TypeName>;
 concrete productions top::TypeName_c
@@ -161,7 +175,7 @@ concrete productions top::Expr_c
  -}
 closed nonterminal InitialFunctionDefinition_c with location, ast<ast:FunctionDecl>, givenStmt;
 concrete productions top::InitialFunctionDefinition_c
-| ds::DeclarationSpecifiers_c  d::Declarator_c  l::DeclarationList_c
+| ds::DeclarationSpecifiers_c  d::FunctionDeclarator_c  l::DeclarationList_c
     {
       ds.givenQualifiers = ds.typeQualifiers;
       d.givenType = ast:baseTypeExpr();
@@ -180,7 +194,7 @@ concrete productions top::InitialFunctionDefinition_c
       -- parameters, and close it after the brace.
       context = lh:beginFunctionScope(d.declaredIdent, d.declaredParamIdents, context);
     }
-| d::Declarator_c  l::DeclarationList_c
+| d::FunctionDeclarator_c  l::DeclarationList_c
     {
       d.givenType = ast:baseTypeExpr();
       local bt :: ast:BaseTypeExpr =
@@ -195,6 +209,40 @@ concrete productions top::InitialFunctionDefinition_c
       -- (i.e. LALR conflicts)
       context = lh:beginFunctionScope(d.declaredIdent, d.declaredParamIdents, context);
     }
+--| ds::DeclarationSpecifiers_c  d::Declarator_c  q::OptTypeQualifierList_c
+--    {
+--      ds.givenQualifiers = ds.typeQualifiers;
+--      d.givenType = ast:baseTypeExpr();
+--      
+--      local bt :: ast:BaseTypeExpr =
+--        ast:figureOutTypeFromSpecifiers(ds.location, ds.typeQualifiers, ds.preTypeSpecifiers, ds.realTypeSpecifiers, ds.mutateTypeSpecifiers);
+--
+--      local specialSpecifiers :: ast:SpecialSpecifiers =
+--        foldr(ast:consSpecialSpecifier, ast:nilSpecialSpecifier(), ds.specialSpecifiers);
+--
+--      top.ast = 
+--        ast:functionDecl(ds.storageClass, specialSpecifiers, bt, d.ast, d.declaredIdent, ds.attributes, ast:nilDecl(), top.givenStmt);
+--    }
+--    action {
+--      -- Function are annoying because we have to open a scope, then add the
+--      -- parameters, and close it after the brace.
+--      context = lh:beginFunctionScope(d.declaredIdent, d.declaredParamIdents, context);
+--    }
+--| d::Declarator_c  q::OptTypeQualifierList_c
+--    {
+--      d.givenType = ast:baseTypeExpr();
+--      local bt :: ast:BaseTypeExpr =
+--        ast:figureOutTypeFromSpecifiers(d.location, ast:nilQualifier(), [], [], []);
+--
+--      top.ast = 
+--        ast:functionDecl([], ast:nilSpecialSpecifier(), bt, d.ast, d.declaredIdent, ast:nilAttribute(), ast:nilDecl(), top.givenStmt);
+--    }
+--    action {
+--      -- Unfortunate duplication. This production is necessary for K&R compatibility
+--      -- We can't make it a proper optional nonterminal, since that requires a reduce far too early.
+--      -- (i.e. LALR conflicts)
+--      context = lh:beginFunctionScope(d.declaredIdent, d.declaredParamIdents, context);
+--    }
 
 
 synthesized attribute isDeclListEmpty :: Boolean;
@@ -317,11 +365,103 @@ concrete productions top::DirectDeclarator_c
       top.ast = dd.ast;
     }
 
+closed nonterminal DirectFunctionDeclarator_c with location, declaredIdent, declaredParamIdents, ast<ast:TypeModifierExpr>, givenType;
+concrete productions top::DirectFunctionDeclarator_c
+| id::Identifier_c 
+    { top.declaredIdent = id.ast;
+      top.declaredParamIdents = nothing();
+      top.ast = top.givenType;
+    }
+| '(' d::FunctionDeclarator_c ')'
+    { top.declaredIdent = d.declaredIdent;
+      top.declaredParamIdents = d.declaredParamIdents;
+      d.givenType = ast:parenTypeExpr(top.givenType);
+      top.ast = d.ast;
+    }
+| dd::DirectFunctionDeclarator_c '[' q::TypeQualifierList_c e::AssignExpr_c ']'
+    { top.declaredIdent = dd.declaredIdent;
+      top.declaredParamIdents = dd.declaredParamIdents;
+      dd.givenType = ast:arrayTypeExprWithExpr(top.givenType, q.typeQualifiers, ast:normalArraySize(), e.ast);
+      top.ast = dd.ast;
+    }
+| dd::DirectFunctionDeclarator_c '[' e::AssignExpr_c ']'
+    { top.declaredIdent = dd.declaredIdent;
+      top.declaredParamIdents = dd.declaredParamIdents;
+      dd.givenType = ast:arrayTypeExprWithExpr(top.givenType, ast:nilQualifier(), ast:normalArraySize(), e.ast);
+      top.ast = dd.ast;
+    }
+| dd::DirectFunctionDeclarator_c '[' q::TypeQualifierList_c ']'
+    { top.declaredIdent = dd.declaredIdent;
+      top.declaredParamIdents = dd.declaredParamIdents;
+      dd.givenType = ast:arrayTypeExprWithoutExpr(top.givenType, q.typeQualifiers, ast:normalArraySize());
+      top.ast = dd.ast;
+    }
+| dd::DirectFunctionDeclarator_c '[' ']'
+    { top.declaredIdent = dd.declaredIdent;
+      top.declaredParamIdents = dd.declaredParamIdents;
+      dd.givenType = ast:arrayTypeExprWithoutExpr(top.givenType, ast:nilQualifier(), ast:normalArraySize());
+      top.ast = dd.ast;
+    }
+| dd::DirectFunctionDeclarator_c '[' 'static' q::TypeQualifierList_c e::AssignExpr_c ']'
+    { top.declaredIdent = dd.declaredIdent;
+      top.declaredParamIdents = dd.declaredParamIdents;
+      dd.givenType = ast:arrayTypeExprWithExpr(top.givenType, q.typeQualifiers, ast:staticArraySize(), e.ast);
+      top.ast = dd.ast;
+    }
+| dd::DirectFunctionDeclarator_c '[' 'static' e::AssignExpr_c ']'
+    { top.declaredIdent = dd.declaredIdent;
+      top.declaredParamIdents = dd.declaredParamIdents;
+      dd.givenType = ast:arrayTypeExprWithExpr(top.givenType, ast:nilQualifier(), ast:staticArraySize(), e.ast);
+      top.ast = dd.ast;
+    }
+| dd::DirectFunctionDeclarator_c '[' q::TypeQualifierList_c 'static' e::AssignExpr_c ']'
+    { top.declaredIdent = dd.declaredIdent;
+      top.declaredParamIdents = dd.declaredParamIdents;
+      dd.givenType = ast:arrayTypeExprWithExpr(top.givenType, q.typeQualifiers, ast:staticArraySize(), e.ast);
+      top.ast = dd.ast;
+    }
+| dd::DirectFunctionDeclarator_c '[' q::TypeQualifierList_c '*' ']'
+    { top.declaredIdent = dd.declaredIdent;
+      top.declaredParamIdents = dd.declaredParamIdents;
+      dd.givenType = ast:arrayTypeExprWithoutExpr(top.givenType, q.typeQualifiers, ast:starArraySize());
+      top.ast = dd.ast;
+    }
+| dd::DirectFunctionDeclarator_c '[' '*' ']'
+    { top.declaredIdent = dd.declaredIdent;
+      top.declaredParamIdents = dd.declaredParamIdents;
+      dd.givenType = ast:arrayTypeExprWithoutExpr(top.givenType, ast:nilQualifier(), ast:starArraySize());
+      top.ast = dd.ast;
+    }
+| dd::DirectFunctionDeclarator_c '(' ptl::ParameterTypeList_c ')'
+    { top.declaredIdent = dd.declaredIdent;
+      top.declaredParamIdents = -- use the inner one if it exists!
+        orElse(dd.declaredParamIdents, just(ptl.declaredIdents));
+      dd.givenType = ast:functionTypeExprWithArgs(top.givenType,
+        ast:foldParameterDecl(ptl.ast), ptl.isVariadic, ast:nilQualifier());
+      top.ast = dd.ast;
+    }
+| dd::DirectFunctionDeclarator_c '(' idl::IdentifierList_c ')'
+    { top.declaredIdent = dd.declaredIdent;
+      top.declaredParamIdents =
+        orElse(dd.declaredParamIdents, just(idl.declaredIdents));
+      dd.givenType = ast:functionTypeExprWithoutArgs(top.givenType,
+        idl.declaredIdents, ast:nilQualifier());
+      top.ast = dd.ast;
+    }
+| dd::DirectFunctionDeclarator_c '(' ')'
+    { top.declaredIdent = dd.declaredIdent;
+      top.declaredParamIdents =
+        orElse(dd.declaredParamIdents, just([]));
+      dd.givenType = ast:functionTypeExprWithoutArgs(top.givenType, [],
+        ast:nilQualifier());
+      top.ast = dd.ast;
+    }
+
 closed nonterminal OptTypeQualifierList_c with location, typeQualifiers;
 concrete productions top::OptTypeQualifierList_c
 |
     { top.typeQualifiers = ast:nilQualifier(); }
-| '<' q::TypeQualifierList_c '>'
+| q::TypeQualifierList_c
     { top.typeQualifiers = q.typeQualifiers; }
```

krame505: Pretty close, just a few simple things:
* `DirectFunctionDeclarator_c` productions should recurse to `DirectDeclarator_c` instead - since we only care about *terminating* the function definition with an `OptTypeQualfiersList_c`, it is OK for these to occur in other places in the declarator (for example, as type qualifiers on a function pointer return type.)  
* In the new `InitialFunctionDefinition_c` productions, presumably these should be using `FunctionDeclarator_c` and followed by a non-optional `TypeQualifiersList_c`?  I'm not sure how this is isn't ambiguous otherwise - does this compile for you?  Also FYI you are currently doing nothing with those qualifiers when constructing the abstract syntax.  
* If we are going to be changing the grammar this significantly from what is specified in the C standard, we should add a comment clearly explaining what is going on here and why we needed to make this change.

TravisCarlson: No, it didn't compile for me, but it still doesn't after making those changes. There are 44 reduce/reduce conflicts between `DirectDeclarator_c` and `DirectFunctionDeclarator_c`.

krame505: Could you push those changes to I can take a look?

TravisCarlson: OK, it's pushed.

krame505:
I figured out an easy way of fixing the conflict with `_Atomic`, so I think we can move ahead on this with the refactored grammar.  I'll go ahead and add the missing equations on concrete syntax, then this should be good to merge.

PR #125, opened 2018-11-08
Implement wrapper productions for decorated forwarding

krame505:
This is something Eric and I discussed, similar to what we do in Silver.  See https://github.com/melt-umn/silver/issues/86 for reference.  

Also fixes some random bugs related to the env that I found.


krame505:
This should be ready to merge, but since this is a "non-trivial" change to ableC (adding new productions), I would like someone to give it a quick review first, maybe @TravisCarlson?  Thanks.

PR #128, opened 2018-11-24
Implement deferred declarations

krame505:
As (previously) existing, the template extension had a bug where a struct is instantiated with type parameters such that some of the struct members are incomplete types.  For example

```C++
template<a>
struct foo {
  a val;
};

inst foo<struct bar> *f;

struct bar {
  int a;
}
```

Here the definition of the instantiation of `foo` would be placed before the declaration of `f` (and definition of `bar`), thus defining a struct containing an incomplete member type.  However, since `f` is a pointer, its struct type *can* be incomplete at this point without any ill effects (this is legal in C++.)

To remedy this, I have modified ableC with a new host production `deferredDecl :: (Decl ::= String Decl)`, wrapping a refId and another declaration.  The semantics of this are that during the `host` transformation, the Decl is inserted at the current level of declarations after the first Decl defining the refId.  This is implemented using a new env scope to hold current deferred decls, and inspecting the defs of every Decl that can contain a type expression, checking if the refId has been defined.  If the refId is never defined, then the Decl is not included in the host tree and the type remains incomplete.  

Note that this transformation is part of `host` and not `lifted` (as might be expected), since extensions can introduce new refIds using the `defsDecl` production that is transformed away in `host`.  In the future, we may want to do some major re-thinking of these transformations, possibly swapping the order of `host` and `lifted` or combining them into one transformation.  

This pull request includes changes also in #127, so it should be merged after that PR.


ericvanwyk:
Can this be documented somewhere?  Maybe in the ableC tutorials?

krame505:
Yeah, when I have time I'll modify the tuple tutorial example to use this.  It has the same issue as templates.

krame505:
I updated the tutorial example.  Also the downstream extensions are all building now, so this is pretty much ready to merge if somebody wants to takes a stab at reviewing this.

PR #129, opened 2018-11-24
Implement attribute to transform away noncanonical types

krame505:
This is needed when an extension uses a type that may have been computed by decorating e.g. a `typedefTypeExpr` outside of the context in which the typedef is valid, but knowing that the underlying type in the typedef is still valid.  This is kind of an unusual corner case, but comes up when trying to statically check type-parametric extension constructs (i.e. checking at definition time and not at the time of use) and due to the introduction of `extType`, can only be done as a part of the host.  This essentially amounts to another default-less functor attribute that all extension types should propagate.  

This PR include changes from #127 and #128, and should be merged after them.


ericvanwyk:
ok with me.

PR #130, opened 2018-11-27
 Workaround for issue with struct tag definitions in inner scopes

krame505:
Minor change, once #128 and #129 are merged.  This fixes a bug with the `decls` attribute on `tagReferenceTypeExpr`.  A plain type expression tag reference of the form `struct foo;` has the effect of overriding the declaration of `foo` in an outer scope, while a declaration involving a tag reference will not do so (i.e. `typedef struct foo _unused;`.)  An extension that introduces a parameterized type (e.g. templates) would presumably want the semantics of a tag reference as a type parameter *not* corresponding to a forward declaration of a new struct (i.e. `inst foo<struct bar>` does *not* correspond to declaring a new struct `bar` overriding a global definition of `struct bar`), so inserting the `decls` from the instantiation type parameters should correspond to only declaring the struct if there is no declaration visible in *any* scope.


krame505:
Ok, this build is passing as well now so I'm just waiting on approval to merge this.

PR #132, opened 2019-01-24
Make assignment to an array index, function call or member directly overloadable

### LINKED ISSUE #131: Array Subscript assign in new overloading
With the old overloading method, it was possible to overload array subscript assignments (so A[x] = ... ). I was wondering if there's a method of doing this with the new overloading? For the tensor algebra extension, this is relevant because there is different behavior if we are assigning a value to a specific index of a tensor, instead of accessing a value out of a tensor. Thanks for the help.

krame505:
Fixes #131.


krame505:
Just FYI, I haven't actually tested anything using these overloads, so let me know if you run into any bugs.

counc009:
I have the tensor algebra extension working with it, so it seems to be working fine, but I'll keep my eye out for problems.

PR #133, opened 2019-01-29
Better error handling for disambiguation

krame505:
Avoid raising an immediate syntax error in case of lookup failure when `TypeName_t` isn't in the shiftable set.  In this case if parsing succeeds there will always be a more helpful semantic error message.


ericvanwyk:
Can you say what you did to achieve this?  Do we return a variable name or type name terminal instead of nothing?  Bring it up in the meeting tomorrow.  Thanks.

ericvanwyk:
OK, so when you "arbitrarily" pick one from the shiftable set you can be sure it will cause a type or name-lookup error during attribute evaluation?  That sounds good to me.

krame505:
Yes, that is correct.  I discuss this in the writeup: https://github.umn.edu/melt/lrk-scratch/blob/master/disambiguation_classes.md

krame505:
@ericvanwyk did you still want to chat more about this before it is merged?  If not, could you go to "Files changed -> Review changes" and click Approve?

PR #134, opened 2019-01-30
Allow '{' terminal to follow type expressions

krame505:
This is allowed in C++, for example in a lambda expression `[](void) -> int { ... }`, and so it is desirable to allow extensions to do the same.  
Some annoying hacks were required due to the ambiguity of having a struct reference in the type expression, for example `[](void) -> struct foo { ... }` - it is highly unlikely someone would wish to declare a struct inside the return type of a lambda, so we wish to disallow this.  This was accomplished by making a new '{' terminal for use in struct/enum/union declarations, and clever use of parser attributes to control the disambiguation between these.

### DIFF
--- a/grammars/edu.umn.cs.melt.ableC/concretesyntax/Declarations.sv
+++ b/grammars/edu.umn.cs.melt.ableC/concretesyntax/Declarations.sv
@@ -164,6 +164,10 @@ concrete productions top::Expr_c
     { top.ast = ast:errorExpr ( [ err (top.location, "Internal Error. " ++
         "Placeholder for TypeNames_c should not appear in the tree.") ],
         location=top.location ) ; }
+| 'TypeNames_NEVER_t!!!nevernever1234567890' TypeNames_c AllowSEUDecl_t
+    { top.ast = ast:errorExpr ( [ err (top.location, "Internal Error. " ++
+        "Placeholder for TypeNames_c should not appear in the tree.") ],
+        location=top.location ) ; }
 | 'Names_NEVER_t!!!nevernever1234567890' Names_c ')'
     { top.ast = ast:errorExpr ( [ err (top.location, "Internal Error. " ++
         "Placeholder for Names_c should not appear in the tree.") ],
--- a/grammars/edu.umn.cs.melt.ableC/concretesyntax/Declarations.sv
+++ b/grammars/edu.umn.cs.melt.ableC/concretesyntax/Declarations.sv
@@ -139,7 +139,7 @@ concrete productions top::Names_c
 | 
   { top.ast = []; }
 
--- Ugly hack to add things to the follow set TypeNames_c and Names_c
+-- Ugly hack to add things to the follow sets of TypeNames_c and Names_c
 -- We set this to include what is allowed by C++ for extensions to use
 terminal TypeNames_NEVER_t 'TypeNames_NEVER_t!!!nevernever1234567890';
 terminal Names_NEVER_t 'Names_NEVER_t!!!nevernever1234567890';
--- a/grammars/edu.umn.cs.melt.ableC/concretesyntax/Declarations.sv
+++ b/grammars/edu.umn.cs.melt.ableC/concretesyntax/Declarations.sv
@@ -148,6 +148,10 @@ concrete productions top::Expr_c
     { top.ast = ast:errorExpr ( [ err (top.location, "Internal Error. " ++
         "Placeholder for TypeNames_c should not appear in the tree.") ],
         location=top.location ) ; }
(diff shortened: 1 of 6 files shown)


krame505:
I would like to merge this soon, unless there are any objections.

ericvanwyk:
Let's talk about this tomorrow in the group meeting.  Be prepared to show us what you did before and what is done now with this proposed change.  If we accept it (and I assume we will) we will want to put these grammar hacks that are not really part of the host language but are put there to support expected future extensions in a place where it is obvious what they are for.

krame505:
Sounds good to me.  FWIW, this is the first time I have actually run into something doable with LR(1) that isn't directly possible with LALR(1) parsing, due to the merging of parse states with different lookahead causing a conflict that wouldn't have happened otherwise.

ericvanwyk:
that is interesting - bring it up tomorrow.

grammars/edu.umn.cs.melt.ableC/concretesyntax/Terminals.sv
```diff
@@ -221,8 +221,28 @@ terminal RParen_t      ')' precedence = 1, association = left; -- evidently, par
 terminal LBracket_t    '[';
 terminal RBracket_t    ']';
 terminal LCurly_t      '{'  action { context = head(context) :: context; };
+terminal TypeLCurly_t  /{/  action { context = head(context) :: context; }; -- { used in types, e.g. struct {...
```
ericvanwyk: Why is this between the traditional curly bracket declarations.  It is now separated from the comment below that describes it.

grammars/edu.umn.cs.melt.ableC/concretesyntax/Declarations.sv
```diff
@@ -148,6 +148,10 @@ concrete productions top::Expr_c
     { top.ast = ast:errorExpr ( [ err (top.location, "Internal Error. " ++
         "Placeholder for TypeNames_c should not appear in the tree.") ],
         location=top.location ) ; }
+| 'TypeNames_NEVER_t!!!nevernever1234567890' TypeNames_c '{'
```
ericvanwyk: Does the comment on line 142 mean 'to the follow sets OF TypeNames_c ...' ?  Do we really need two (lines 145 and 146) "unscannable" terminals?  Is one not sufficient?

grammars/edu.umn.cs.melt.ableC/concretesyntax/Terminals.sv
krame505: Good point, I will move it.

grammars/edu.umn.cs.melt.ableC/concretesyntax/Declarations.sv
krame505: Yes, that is a typo in the comment.
We need 2 seperate "unscannable" terminals to avoid a lexical ambiguity.  It would be really nice to give Copper some support for this - maybe just as simple as introducing a form of regex that will never match or be ambiguous with any other regex?

grammars/edu.umn.cs.melt.ableC/concretesyntax/Terminals.sv
```diff
@@ -273,7 +274,7 @@ terminal LEFT_OP       '<<'    lexer classes {Csymbol};
 
 -- Numerical operators
 terminal Minus_t       '-'  precedence = 5, association = left, lexer classes {Csymbol}; -- negative
-terminal Plus_t        '+'  precedence = 5, association = left,  lexer classes {Csymbol}; -- positive
+terminal Plus_t        '+'  precedence = 5, association = left, lexer classes {Csymbol}; -- positive
```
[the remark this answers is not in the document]
ericvanwyk: OK.  What does associativity have to do with unary operators?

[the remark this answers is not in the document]
krame505: After looking at the grammar again to remind myself, these terminals are *also* the binary addition/subtraction operators, as well as being unary operators.  Even so, due to the way we structure the grammar, these precedence and associativity values aren't actually doing anything.  I suppose it makes sense to have them as a reminder or if an extension where to use them for some sort of DSL expression.

ericvanwyk:
If these have no effect we should remove them.

krame505:
OK, but that is a seperate change so I would prefer to do it in another pull request.  I'll open an issue.

ericvanwyk:
Makes sense to do another issue and PR - thanks.

krame505:
@ericvanwyk reminder to approve this when you get a chance.

PR #135, opened 2019-02-01
Change asm concrete syntax to accept multi-line strings

counc009:
I had mentioned a problem with multi-line strings earlier today. It turns out that the problem was just with how ableC constructed the asm construct, it used StringLiteral_t instead of StringLiteral_c which allows for multi-line strings. This brings ableC's implementation in line with GCC's implementation, and makes using this feature in ableC reasonable.


krame505:
Should be fixed, it was actually some a latent bugs related to label scoping I uncovered in the develop branch of a couple downstream extensions.  Github should let you merge as soon as it's done building, I think.

PR #144, opened 2019-04-22
Re-implement substitution using reflection instead of as a functor attribute.  

### LINKED ISSUE #121: Substitution, as currently conceived, is interfering
I ran into this case when thinking about how to implement templated algebraic data types.  The issue essentially stems from that we want to perform substitution on ASTs to perform rewriting "before" forwarding occurs, without using the env.  However, changing names that are children of an extension productions can affect the forwards-to tree in ways beyond just changing places where the same name was directly used - e.g. an extension may create other names derived from the original (substituted) name in the tree.  For example, if we have an extension that defines a production
```
abstract production fooDecl
top::Decl ::= id::Name
{
  -- propagate substituted;
  forwards to
    ableC_Decl {
      typedef int $name{"foo_" ++ id.name};
    };
}
```
then simply writing `propagate substituted;` would actually be interfering!  Performing a name substitution of "x" for "x1" and then computing `host` on the tree `fooDecl(name("x"))` would yield the tree `typedefDecls(intType(), declarator(..., name("x"), ...))` if evaluating via the forward, but with the `propagate substituted` equation, we would get `typedefDecls(intType(), declarator(..., name("x1"), ...))`, which although is the value that we "want" to get, is actually considered "wrong" by our current notion of non-interference.  

This may not seem to be an issue at first (just make sure everyone writes `propagate substituted;` everywhere!), but this actually can come back to bite us in several ways.  First of all, we have no theoretical framework (at present) for actually requiring extension writers to include such equations, or for reasoning about the behaviour we will get from composing such extensions.  According to our current MWDA and non-interference rules, the writer of `fooDecl` is *doing nothing wrong* by leaving out an equation for `substituted`.  Doing so could result in some rather unexpected behaviour that is difficult could be difficult to diagnose.  

Going forward, we should give some thought to alternative ways of achieving what we want to do with `substituted`.  What we *really* want (for things such as the template extension) is to have some way of performing "syntactic" substitutions for all occurrences of certain subtrees on the *undecorated* tree.  This could be done using reflection/reification, but we need to think carefully about the semantics of this.  Presumably we would want to avoid having extensions directly using reflection on host trees in computing their forward (as reasoning about this would get out of hand), but we could potentially think of having the host provide some set of built-in constructs to perform various types of substitutions, with specific extra semantic meanings given to various host productions not directly specifiable via attributes - i.e. the specific production `name`, when a `nameSubstitution` is applied with a matching name, will be matched and renamed at any point in the tree, etc.

krame505:
Fixes #121.

Currently the build is failing as a few of the downstream extensions still need updating.  

Also more thought may be needed as to the best way of handling substitution on "irregular" productions that don't directly correspond to concrete syntax, such as `directTypeExpr` and the decorated wrapper productions.  Currently there are some hacky workarounds in place to allow special handling for these productions.


krame505:
All the downstream extensions and the build is passing, so this should be ready to merge.  @ericvanwyk can you take a look?

krame505:
At least 2200 LOC within ableC, but there are at a bunch more in the extensions where we got rid of `propagate substituted;`.

ericvanwyk:
It would be good to have the precise number.  For this and the other code-reductions that were done with the reflection extension.

Issue #145, opened 2019-05-12
Type Qualifiers  don't work with typedef types


counc009:
I've run into a problem with type qualifiers in typedef types. I think the issue is simply that the qualifiers aren't applied properly to the canonicalType of the NoncanonicalType nonterminal by the typedefType production. This problem causes qualifiers to not cause modifiers to be applied to qualified typedef types.

counc009:
As a very simple example of this issue, it can be seen to occur on type checking. Looking at the type-qualifiers tutorial, try the following piece of code:
> typedef int type_t;
>
> int main() {
> &nbsp;&nbsp;&nbsp;&nbsp;tainted type_t a = (tainted type_t) 1;
> &nbsp;&nbsp;&nbsp;&nbsp;untainted type_t b = (untainted type_t) 2;
> &nbsp;&nbsp;&nbsp;&nbsp;a = b;
> &nbsp;&nbsp;&nbsp;&nbsp;b = a;
> }

This should cause a compilation error, but doesn't. For instance, replace `type_t` with int in the code above and we do get a compilation error as expected.

counc009:
Resolved by merge of pull request #146.

PR #146, opened 2019-05-12
Modify typedef production so canonicaltype includes qualifiers

counc009:
Fixes error described in #145. Adds qualifiers to canonicaltype so that modifications caused by qualifiers affect the variable.

### DIFF
--- a/grammars/edu.umn.cs.melt.ableC/abstractsyntax/host/Types.sv
+++ b/grammars/edu.umn.cs.melt.ableC/abstractsyntax/host/Types.sv
@@ -805,7 +805,7 @@ top::NoncanonicalType ::= q::Qualifiers  n::String  resolved::Type
     noncanonicalType(
       typedefType(foldQualifier(top.addedTypeQualifiers ++ q.qualifiers), n, resolved));
 
-  top.canonicalType = addQualifiers(q.qualifiers, resolved);
+  top.canonicalType = resolved;
 }
 
 {-- GCC typeof type expression -}
--- a/grammars/edu.umn.cs.melt.ableC/abstractsyntax/host/Types.sv
+++ b/grammars/edu.umn.cs.melt.ableC/abstractsyntax/host/Types.sv
@@ -803,7 +803,10 @@ top::NoncanonicalType ::= q::Qualifiers  n::String  resolved::Type
   top.typeModifierExpr = baseTypeExpr();
   top.withTypeQualifiers =
     noncanonicalType(
-      typedefType(foldQualifier(top.addedTypeQualifiers ++ q.qualifiers), n, resolved));
+      typedefType(
+        foldQualifier(top.addedTypeQualifiers ++ q.qualifiers),
+        n, resolved.withTypeQualifiers));
+  resolved.addedTypeQualifiers = top.addedTypeQualifiers;
 
   top.canonicalType = resolved;
 }
@@ -821,7 +824,10 @@ top::NoncanonicalType ::= q::Qualifiers  resolved::Type
   top.typeModifierExpr = baseTypeExpr();
   top.withTypeQualifiers =
     noncanonicalType(
-      typeofType(foldQualifier(top.addedTypeQualifiers ++ q.qualifiers), resolved));
+      typeofType(
+        foldQualifier(top.addedTypeQualifiers ++ q.qualifiers),
+        resolved.withTypeQualifiers));
+  resolved.addedTypeQualifiers = top.addedTypeQualifiers;
 }
 
 function filterExtensionQualifiers
--- a/grammars/edu.umn.cs.melt.ableC/abstractsyntax/host/Types.sv
+++ b/grammars/edu.umn.cs.melt.ableC/abstractsyntax/host/Types.sv
@@ -805,7 +805,7 @@ top::NoncanonicalType ::= q::Qualifiers  n::String  resolved::Type
     noncanonicalType(
       typedefType(foldQualifier(top.addedTypeQualifiers ++ q.qualifiers), n, resolved));
 
-  top.canonicalType = resolved;
+  top.canonicalType = addQualifiers(q.qualifiers, resolved);
 }
 
 {-- GCC typeof type expression -}


krame505:
I don't think that's the source of the problem - can you update the issue with an example?  `resolved` in `typedefType` should already have had the qualifiers added, see the equation for `typerep` in `typedefTypeExpr`.

counc009:
Okay, I just checked and it does seem to have qualifiers, but for some reason it's not working properly. I don't have a great example, what I'm testing with is an odd hacked together extension that currently relies on things from several branches of ableC. I'll do my best to give an example and show how my code works, but if it would be more helpful I can also push this onto a repo on the umn github.

I've introduced a type qualifier `negate`, and given it qualifiers so that when a variable qualified with it is used in a post-increment expression the variable is transformed into the expression __a[0] (where __a is an int array I declare earlier in my code). So, as a test I have
> struct test { int x; }
> typedef int type_t;
> int main() {
> &nbsp;&nbsp;&nbsp;&nbsp;int __a[4] = {1, 2, 3, 4};
> &nbsp;&nbsp;&nbsp;&nbsp;int negate a;
> &nbsp;&nbsp;&nbsp;&nbsp;a++;
> &nbsp;&nbsp;&nbsp;&nbsp;struct test negate b;
> &nbsp;&nbsp;&nbsp;&nbsp;b++;
> &nbsp;&nbsp;&nbsp;&nbsp;type_t negate c;
> &nbsp;&nbsp;&nbsp;&nbsp;c++;
> }

The generated code it produces is
> struct test { int x; }
> typedef int type_t;
> int main() {
> &nbsp;&nbsp;&nbsp;&nbsp;int __a[4] = {1, 2, 3, 4};
> &nbsp;&nbsp;&nbsp;&nbsp;int a;
> &nbsp;&nbsp;&nbsp;&nbsp;(__a[0])++;
> &nbsp;&nbsp;&nbsp;&nbsp;struct test b;
> &nbsp;&nbsp;&nbsp;&nbsp;(__a[0])++;
> &nbsp;&nbsp;&nbsp;&nbsp;type_t c;
> &nbsp;&nbsp;&nbsp;&nbsp;c++;
> }

I've also added a construct in this extension for examining types, that shows the unparse of the expression's type. It shows that the qualifier is provided to the `noncanonicalType` production, but is not getting placed onto `resolved`. If the qualifier is being provided to the `typedefTypeExpr` I agree it should be on `resolved`, so this may suggest a problem somewhere else in ableC that is causing the qualifier problem.
(As a note, it doesn't seem to be related to where the qualifier is provided. If I use `negate type_t c` the same problem occurs)

counc009:
I ran into this issue testing my openmp extension on the umn repo together with the Cilk extension. If it would help I could also post that code.

krame505:
Ok, I'm still not sure I entirely understand what you are trying to do, but are you sure that using type qualifiers is really the best approach?  

If you could try to come up with a minimal example that you think illustrates a bug in ableC, that would be helpful.

counc009:
In the OpenMP extension we need to lift the parallel loop body into a function. However, because OpenMP uses a shared memory model we can't just lift the loop body as is, we must modify it so that all the threads running the function access the same memory addresses (so we can't use local variables / arguments, and without entirely modifying all the code outside of the loop we can't use global variables). My first attempt to do this relied on the substitution mechanism, substituting the variable name with an appropriate Expr for accessing the variable using an array of pointers. However, this had problems since the it seems the substitution occurs before forwarding. More problematic, though, was also that it entirely broke in the case of type shadowing within the loop. For example:
> int x = 7;
> #pragma omp parallel for
> for(int i = 0; i < 10; i++) {
> &nbsp;&nbsp;&nbsp;&nbsp;printf("%d\n", x);
> &nbsp;&nbsp;&nbsp;&nbsp;double x = 4.2;
> &nbsp;&nbsp;&nbsp;&nbsp;printf("%f\n", x);
> }

This would cause problems because using substitution, every occurrence of x will be substituted, even after we declare a new x that isn't shared. On Eric's suggestion I looked at using type qualifiers for this problem, and it seems to work as desired, except for the issue with typedefs. 

I'll try to post a simple example of an extension that demonstrates this bug as soon as possible.

krame505:
I think I might understand what you are trying to do, but why do you need an array?  Couldn't you just do what the closure extension is doing to capture variables, or even use the closure extension itself?

counc009:
As a very simple example of this issue, it also is a problem in type checking. Looking at the type-qualifiers tutorial, try the following piece of example:
> typedef int type_t;
>
> int main() {
> &nbsp;&nbsp;&nbsp;&nbsp;tainted type_t a = (tainted type_t) 1;
> &nbsp;&nbsp;&nbsp;&nbsp;untainted type_t b = (untainted type_t) 2;
> &nbsp;&nbsp;&nbsp;&nbsp;a = b;
> &nbsp;&nbsp;&nbsp;&nbsp;b = a;
> }

This should cause a compilation error, but doesn't. For instance, replace `type_t` with int in the code above and we do get a compilation error.

As for the array, we could get away with a struct I guess, but we do have to pass pointers rather than just passing values like the closure extension.

krame505:
Yes, I agree that this is a bug - could you go update #145 with this example?  

If I understand correctly, the issue with passing things by value in a struct and unpacking them back to variables with the same names when inside the function is that assignments to the variables would be lost when the function returns.  Could you just simply write the current values of all local variables back into the struct at the end of the function, and then after the function returns at the call site, update all captured variables with the values from the struct?

counc009:
I posted the example to the bug report.

The problem with copying values to and from a struct is that it doesn't actually emulate OpenMP's memory model. OpenMP really shared between threads, so since the lifted function will be performed by multiple threads in parallel they must make their updates directly to memory. For example:
> int main() {
> &nbsp;&nbsp;&nbsp;&nbsp;int x = 0;
> &nbsp;&nbsp;&nbsp;&nbsp;#pragma omp parallel for num_threads(4)
> &nbsp;&nbsp;&nbsp;&nbsp;for(int i = 0; i < 1000; i++) {
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;x++;
> &nbsp;&nbsp;&nbsp;&nbsp;}
> &nbsp;&nbsp;&nbsp;&nbsp;printf("x = %d\n", x);
> }

This results (nondeterministically) in values for x, on my machine varying from near 250 to 1000, so we have to make the updates visible to all threads simultaneously. Also, looking at the generated assembly we can see that the code generated by OpenMP works like what I'm working to implement.

sweetiepiggy:
Here is what I think is going on with the typedef problem.

The equation for `typerep` in `typedefTypeExpr` is
```
top.typerep = noncanonicalType(typedefType(q, name.name, addQualifiers(q.qualifiers, name.valueItem.typerep)));
```

The equation for `qualifiers` in `noncanonicalType` is obtained by forwarding to `canonicalType` on the `typedefType(...)` above. The equation is then `top.canonicalType = resolved;` where `resolved` in this case is `addQualifiers(q.qualifiers, name.valueItem.typerep);`.

In the given example it appears that  `name.valueItem.typerep` is `noncanonicalType(typedefType(q, n, builtinType(...)))`. The `addQualifiers()` call updates the `q`, but when we look at the `qualifiers` attribute on the resulting tree we again look at `canonicalType` on the `typedefType(...)` and ignore the new `q`, finally getting the nil `qualifiers` on `builtinType(...)`.

In other words, we don't have the following property:

forall q in qs . member(q, `addQualifiers(qs, noncanonicalType(...)).qualifiers`)

TravisCarlson:
Here is what I think is going on with the typedef problem.

The equation for `typerep` in `typedefTypeExpr` is
```
top.typerep = noncanonicalType(typedefType(q, name.name, addQualifiers(q.qualifiers, name.valueItem.typerep)));
```

The equation for `qualifiers` in `noncanonicalType` is obtained by forwarding to `canonicalType` on the `typedefType(...)` above. The equation is then `top.canonicalType = resolved;` where `resolved` in this case is `addQualifiers(q.qualifiers, name.valueItem.typerep);`.

In the given example it appears that  `name.valueItem.typerep` is `noncanonicalType(typedefType(q, n, builtinType(...)))`. The `addQualifiers()` call updates the `q`, but when we look at the `qualifiers` attribute on the resulting tree we again look at `canonicalType` on the `typedefType(...)` and ignore the new `q`, finally getting the nil `qualifiers` on `builtinType(...)`.

In other words, we don't have the following property:

forall q in qs . member(q, `addQualifiers(qs, noncanonicalType(...)).qualifiers`)

krame505:
Thanks, Travis - I see the issue now.  The problem is that when adding qualifiers on a `typedefType` we also need to add them on `resolved`.  We have the same bug with `typeofType`.  I just commited that fix on this branch, @counc009 can you give this a try and let me know if the issue is fixed?  If so I will go ahead and approve to merge.

counc009:
Yes, that seems to have fixed the problems I was seeing. Thanks for the help.

Issue #150, opened 2019-05-19
Function scope lifting and nested functions


counc009:
In our implementation of function scope lifting (see #149), we currently inject definitions into the second outermost scope of the environment. In general, this is correct, but when we have nested functions this is incorrect, as the declarations will be lifted to the outermost scope of the nested function, but will be placed into the environment of the outermost scope of the outer function.
Currently, I don't believe there to be a solution to this, it would likely require modifying how environments are designed in a manner that would allow us to identify what scope is the inner-most function scope.

krame505:
Just to clarify, I think we are currently injecting declarations into the innermost function scope; the issue is that the environment being provided to the `Decls` being lifted only contains declarations from the outermost function scope and above.

PR #153, opened 2019-08-28
Fix #152 by removing functionDecls from functionDecl's defs

### LINKED ISSUE #152: Error with function scope lifting
I've run into a problem with my implementation of function scope lifting. The definition (not the actual definition) gets lifted to a global scope, so if two functions try to inject the same value to function scope only the first function will succeed. 

A demonstration of this can be difficult, but the following silver can be used to produce an extension that will demonstrate the bug.
```
grammar edu:umn:cs:melt:exts:ableC:injectionbug;

imports edu:umn:cs:melt:ableC:concretesyntax;
imports silver:langutil only ast;

marking terminal Inject_t 'inject' lexer classes {Ckeyword};
marking terminal Arrow_t '<-' lexer classes {Ckeyword};

concrete production inject_c
s::Stmt_c ::= 'inject' 'int' n::Identifier_t '<-' e::Expr_c ';'
{
  s.ast = injectInt(n.lexeme, e.ast);
}

marking terminal Env_t 'env' lexer classes {Ckeyword};

concrete production env_c
s::Stmt_c ::= e::Env_t ';'
{
  s.ast = dispEnv(e.location);
}
imports edu:umn:cs:melt:ableC:abstractsyntax:construction;
imports edu:umn:cs:melt:ableC:abstractsyntax:env;
imports edu:umn:cs:melt:ableC:abstractsyntax:host;

abstract production injectInt
s::Stmt ::= n::String v::Expr
{
  forwards to
    declStmt(injectFunctionDeclsDecl(consDecl(maybeValueDecl(
      n, variableDecls(nilStorageClass(), nilAttribute(),
      directTypeExpr(
        builtinType(nilQualifier(), signedType(intType()))),
      consDeclarator(
        declarator(name(n, location=v.location), baseTypeExpr(), nilAttribute(),
        justInitializer(exprInitializer(v))),
        nilDeclarator()))), nilDecl())));
}

abstract production dispEnv
s::Stmt ::= loc::Location
{
  forwards to exprStmt(printEnv(location=loc));
}
```

And then this xc file will show the problem
```C
#include <stdio.h>

int f() {
   inject int x <- 4;
   printf("x = %d\n", x);
}

int main() {
    inject int x <- 3;
    printf("x = %d\n", x);
}
```

With ableC currently, the resulting lifted pp (cleaned for clarity):
```C
int f() {
    signed int x = 4;
    printf("x = %d\n", x);
}

int main() {
    printf("x = %d\n", x);
}
```

counc009:
This error is described in detail in #152.
Using the example given there, this fix now results in proper C code where both declarations are injected as desired. 
The issue was caused by adding the functionDecls into the defs attribute from functionDecl production, so that it appeared to other functions as if the declaration had already been injected into global scope. 

This does also identify a larger problem with using `maybeValueDecl` for function scope lifting, which perhaps we should look into solving as well.
We could perhaps solve that by creating a new function `maybeValueFunctionDecl` (or some such name) and having it check only the non-global environment, though that will perpetuate the bug in #150 about nested functions, even further.

### DIFF
--- a/grammars/edu.umn.cs.melt.ableC/abstractsyntax/host/Decls.sv
+++ b/grammars/edu.umn.cs.melt.ableC/abstractsyntax/host/Decls.sv
@@ -522,11 +522,7 @@ top::FunctionDecl ::= storage::StorageClasses  fnquals::SpecialSpecifiers  bty::
     globalDeclsDefs(mty.globalDecls) ++
     globalDeclsDefs(ds.globalDecls) ++
     globalDeclsDefs(body.globalDecls) ++
-    globalDeclsDefs(fnquals.globalDecls) ++
-    functionDeclsDefs(mty.functionDecls) ++
-    functionDeclsDefs(ds.functionDecls) ++
-    functionDeclsDefs(body.functionDecls) ++
-    functionDeclsDefs(fnquals.functionDecls);
+    globalDeclsDefs(fnquals.globalDecls);
   top.freeVariables :=
     bty.freeVariables ++
     removeDefsFromNames(implicitDefs, mty.freeVariables) ++


krame505:
Fix looks good to me.  
What exactly is the problem with `maybeValueDecl`?

counc009:
My idea was that if you have a global variable (or lifted declaration) for a variable `x`, and then try to lift `x` as a function scope declaration using a `maybeValueDecl` `x` won't be added to the function, since it already exists in the environment.
However, thinking about it, and how its used by extensions, I don't think it's really an issue, since any lifted name should be sufficiently qualified and unique to avoid possible name collisions. If it was a problem, it seems that one of the names would need to be changed.

PR #154, opened 2019-08-31
AbleC 0.1.3 release

krame505:


krame505:
Also if someone could review #151 so it can be merged before the release, that would be nice.

ericvanwyk:
I approved #151.  Let me know when I can get the .tar.gz file for the digital repository.

krame505:
Github will automatically create the tarball when I do the release.  I'll probably make the release tomorrow after I finish working on the paper responses.

ericvanwyk:
Can you do one for silver-ableC too?  I'm not sure if this should be 0.1.0 or make it 0.1.3 to match the version of ableC, thoughts?

krame505:
Previously we haven't been doing formal releases of silver-ableC, we've just been treating it like the other ableC extensions and doing a merge to master/tag corresponding to every ableC release.  But if we want a seperate tarball then an actual release of silver-ableC makes sense.  We probably want to release it through github as 0.1.0, but still add the ableC-0.1.3 tag pointing to the correct snapshot w.r.t. ableC.

ericvanwyk:
Look here:
    https://github.com/melt-umn/silver-ableC/releases
This looks like a release of silver-ableC that is good enough.  It does generate a tarball that has just the contents of the silver-ableC repo.  So doing another one of these seems OK to me.  What do you think?

krame505:
The "releases" tab shows all the tags - I didn't realize this generates a tarball as well.  But for consistency we probably want to also do an "actual" release, which is quick.  

I'm going to wait doing anything with releases until after we finish our work on the paper tomorrow (which is higher priority, right?)

ericvanwyk:
We need to submit it to the DRUM to get the DOI that will go in the final version.  Tomorrow anytime is probably fine, but let's make sure we do it tomorrow.  Thanks.

PR #155, opened 2019-09-01
Add setup injection code to functionDecl production

counc009:
For some of my work on parallelization in ableC, I need extensions to execute setup code before the parallel extensions are used. This generally takes the form of a function call, but must be injected by the parallelization system since this setup depends on the number of threads the extension is given as well as an integer that is unique to the extension.
My idea for doing this is to inject the setup code into the function's body in the functionDecl production if the function's name is "main". I inject this setup code into the host tree, rather than the lifted tree, because I found for my parallelism work that it needs to be able to depend on an extended environment and that putting these definitions in the lifted tree caused the environment to be incorrect. This is also the reason that I have code analyzing the environment, just to ensure that the attribute's flowtype includes the environment.
I'm open to hearing other suggestions if there might be a better way to do this, but this seems to work, and seems like other extensions could use this feature as well.

### DIFF
--- a/grammars/edu.umn.cs.melt.ableC/abstractsyntax/host/Decls.sv
+++ b/grammars/edu.umn.cs.melt.ableC/abstractsyntax/host/Decls.sv
@@ -440,8 +440,19 @@ flowtype FunctionDecl = decorate {env, returnType}, name {}, sourceLocation {};
 abstract production functionDecl
 top::FunctionDecl ::= storage::StorageClasses  fnquals::SpecialSpecifiers  bty::BaseTypeExpr mty::TypeModifierExpr  name::Name  attrs::Attributes  ds::Decls  body::Stmt
 {
-  propagate host;
-  
+  production attribute setups :: [Stmt] with ++;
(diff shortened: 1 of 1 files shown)


krame505:
Tying this specifically to `main` seems a tad inflexible (e.g. how does this work for ableC libraries that have no main function?) but maybe you've thought this through.  

You are right that this injection should happen as part of `host`, but currently your implementation has a bug - `defs` from decorating `setups` should be initially visible in the body.  I don't see a straightforward fix with this approach, though, as you would somehow need to be able to tell that you are in the host tree and _not_ add the defs.  

I would maybe recommend instead moving `setup` to a forwarding `mainSetupStatement` production that just forwards to the setup statements, and introducing this production as an extra statement when constructing the AST for main in the concrete syntax.  

This should wait until after the next ableC release (hopefully today).  Also FYI `host` and `lifted` will be merging at some point soonish.

counc009:
I think I've settled on an approach that requires a bit of extra syntax, but actually will allow the programmer more control, so this isn't necessary for my work. Closing this.

PR #158, opened 2019-12-16
Significant refactoring to ableC's lexer classes

krame505:


krame505:
This fixes a bunch of things that have been bugging me about ableC's lexer classes for a while, and ties up some loose ends with my previous work on disambiguation classes.  There are a bunch of enhancements to lexer classes in Silver in PR # that this depends on.  I'm including Ted on this but no need to comment.  

* Renamed all lexer classes to remove the "C" prefix.  This had been bugging me for a while because extension keywords aren't really "C" keywords, etc. and calling them `AbleCKeywords` would be redundant.  Also the capitalization is weird.  And renaming everything made all the extensions not compile so keeping track of which ones I needed to update was easy :-)
* Renamed terminals to use a consistent naming scheme.  Some where `ALL_CAPS` and some where `CamelCase_t`.  They are now all the latter.
* Split `Ckeyword` into `Keyword` and `Reserved`.  `Keyword`s are things that should be treated as keywords (for highlighting, etc.) and `Reserved` are things that should dominate identifiers.  Extensions can introduce new keywords, but (according to a chat I had with August last year) extensions really shouldn't be using lexical precedence for marking terminals (and thus shouldn't use the `Reserved` lexer class.)  Doing so can break host-language code (if a keyword dominates an identifier of the same name) with no chance to fix it using transparent prefixes or disambiguation functions.  Instead extensions should define keywords that are lexically ambiguous with host terminals and use disambiguation functions.  Note that this means ambiguities between extensions at composition time will also include host terminals, but that is actually desired behavior: for example, if there are two different extensions defining a 'table' marking terminal, we may decide that we want to always prefer the identifier named "table" and require transparent prefixes for both extensions.  See the updated examples at https://github.com/melt-umn/ableC-sample-projects/tree/feature/lexerClasses/using_transparent_prefixes
* Split `Cidentifier` into `Identifier` and `Scoped`.  `Identifier`s are things that should be treated as identifiers (for highlighting, etc.) and `Scoped` are things that should be disambiguated based on `context`.
* Further refinements to using scoped disambiguation for keywords.  This method of disambiguation has several advantages over writing (often numerous) explicit disambiguation functions (for example, dealing with third-party header files that use identifiers with the same name as an extension keyword, when we are unable to edit the code to add transparent prefixes.)  But there are a couple of problems:
  1. Initially adding terminals to `context` is super annoying with the need to do it in a parser action block.
  2. If two extensions introduce the same scoped keyword, the one we get is arbitrarily whichever one got added to `context` last.  Note that with disambiguation as currently implemented, if the lexeme is not found in the context and there is no valid host terminal, we arbitrarily pick something that is valid.  However this is acceptable because in such a case there is guaranteed to be a semantic error that is probably better than the parse error that would otherwise result.  But in the case of two ambiguous keywords in `context`, it is possible that we will pick the wrong one and get a confusing parse error later on, or we pick the right one, and are left with brittle code that only works due to undefined behavior.  

  The first issue is resolved by an enhancement to Silver (PR #) that lets one reference a lexer class in an action block to get the list of its `TerminalId` members.  Extension terminals can specify that they should be disambiguated using lexical scoping and implicitly placed in the outermost scope by specifying the `Global` lexer class.  The `Scoped` disambiguation lexer class, after failing to find the lexeme in the context, checks to see if any of the currently valid terminals are in the `Global` lexer class, and if it finds one, it is returned.  

  As for the second issue, when more than one valid scoped keyword terminal is found, the only consistent thing to do is to disallow all the options and raise a parse error.  The programmer would need to use transparent prefixes for both extensions, or could override the `Scoped` disambiguation by specifying an explicit preference.  See https://github.com/melt-umn/ableC-sample-projects/tree/feature/lexerClasses/using_scoped_keywords more examples of what this looks like in practice.

  A slight drawback of using explicit disambiguation functions/`prefer` clauses in parser specs to override the `Scoped` disambiguation class is that shadowing of the preferred terminal will no longer be respected.  This is probably not a major issue for the programmer as they will already be using transparent prefixes to deal with the ambiguity, but sometimes this kind of ambiguity can occur within an extension.  Ordinarily the extension writer would provide a default resolution using disambiguation functions, but when such ambiguities include identifiers, a different approach is needed to preserve lexical scoping.  This can be done by making both affected keywords `Global` and specifying a preference using the `globalPreferences :: [Pair<[TerminalId] TerminalId>]` parser attribute.  This contains a mapping of conflicting sets of `Global` terminals to the preferred resolution, similar to disambiguation groups.  When the `Scoped` disambiguation class encounters an ambiguity, it checks for a preference resolution before raising a syntax error.  An example of this can be found at https://github.com/melt-umn/ableC-template-algebraic-data-types/blob/feature/lexerClasses/grammars/edu.umn.cs.melt.exts.ableC.templateAlgebraicDataTypes/datatype/concretesyntax/DataType.sv

  I also updated the ableC tutorials with a more detailed discussion of lexer classes and lexical disambiguation: https://github.com/melt-umn/ableC-tutorials/tree/feature/lexerClasses/declarations

Fixes #193, #300.

krame505:
@ericvanwyk could you take a look at this and https://github.com/melt-umn/silver/pull/336 when you get a chance?  Thanks.

PR #167, opened 2020-10-14
Pre-run cpp on all header test cases to avoid dependency on local system headers

krame505:
This fixes an annoying portability issue where sometimes a different set of the failing positive tests fail because the test cases are including system header files that may differ between machines.


remexre (approved):

ericvanwyk:
Can you say a bit more about what the problem was and what the solution was?  What are these header test cases for?  Thanks.

krame505:
Our ableC test suite includes a large number of test cases that are simply system header files.  These should all successfully compile with ableC in theory, however a few of them are currently failing.  However some of these header files include *other* header files that vary between distros/versions/etc. so sometimes a different set of these are failing when the test suite is run on some machines but not others.  Pre-running them all through the C preprocessor eliminates this dependency and makes things consistent.

melt-robot007:
Ok, but if the tests are failing on some machines/distros don't we want to find out what is missing in the ableC spec and fix it?

ericvanwyk:
OK - that we me logged in as the robot0007 account...

krame505:
These are failing for a variety of reasons - some because we don't implement GCC extension properly/completely, some because we haven't gotten to implementing various corners of the C spec and have oversimplifying workarounds in place for now.  This is actually a good project for people who want to dive into working with ableC more, of going through these failing tests and fixing them.

PR #170, opened 2020-11-02
Ignore extension terminals in disambiguation logic when parsing system header files

krame505:
This was suggested by @remexre.  

When parsing system header files, we would like to avoid conflicts with extensions.  This is mostly achieved through my previous work with lexically scoped disambiguation for extension marking terminals, however extension keywords can show up as ambiguous with identifiers in some places that aren't definitions (e.g. struct field names in a header file.)  

It turns out that the C preprocessor tags contain a convenient way to check if we are in a system header file.  So we can actually just turn off disambiguating to extension terminals (at least ones using the `Global` lexer class mechanism) when we are in a system header file (where the obviously shouldn't be any extension terminals.)  The disambiguation logic is otherwise unchanged.

### DIFF
--- a/grammars/edu.umn.cs.melt.ableC/concretesyntax/LexerHack.sv
+++ b/grammars/edu.umn.cs.melt.ableC/concretesyntax/LexerHack.sv
@@ -42,32 +42,30 @@ lexer class Scoped
         else nothing()
       | nothing() -> nothing()
       end;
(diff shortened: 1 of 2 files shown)


krame505:
@ericvanwyk Given that August doesn't think this is a generally applicable enough case to merit special handling in Copper, and the disambiguation function workaround works well enough, are you OK with merging this as-is for now?

ericvanwyk:
Yes, we can merge it for now.

ericvanwyk (approved):

PR #171, opened 2020-11-19
Fix bug in CPP tag location tracking introduced by header disambiguation fix

krame505:
Since all semantic actions for terminals (which include updating the current location) are run when the tokens are accepted by the parser, it turns out that moving the entire action block for `CPP_Location_Tag_t` introduced a bug - the location information was being updated too early, and then being overwritten when the subsequent tokens were accepted by the parser.  This fixes that bug by moving the semantic actions on CPP tags for updating the location back to a regular action block rather than the workaround disambiguation function.


remexre (approved):
I don't grok the patch, but line numbers seem correct?

RIP @ L27

PR #172, opened 2020-11-19
Fix parsing of C11-allowed repeated typedefs

krame505:
This fixes a long-standing bug where C11-allowed repeated `typedef`s were parsed incorrectly.  Recall that declarations are parsed as a sequence of declaration specifiers followed by a list of declarators.  Consider the program
```c
typedef int a;
typedef int a;
```
Currently we correctly in first `typedef` we parse the declaration specifiers `typedef` and `int`, then there is an ambiguity at `a` as it could be a type name as an additional type specifier, or an identifier as a declarator.  Since `a` has no definition we correctly disambiguate it as an identifier/declarator.  

However in parsing the second `typedef`, when we reach `a` we see that it has a definition in the context as a type name, and parse it as another type specifier!  This is a syntactically valid parse which results in a semantic error, as it is illegal for a declaration to contain multiple "real" type specifiers.  

The fix for this is that as soon as we parse a type specifier in a declaration specifier list, ambiguities between type names and identifiers for names that are in the context as types should be resolved as identifiers.  This required a bit of slight grammar restructuring, a new parser attribute, and yet another case in the disambugation class for identifiers.

The downside to this is that there are few cases that formerly were semantic errors that are now parse errors.  For example
```
typedef int a;
typedef int a b;
```
would previously give a semantic error about "multiple type specifiers" on the second typedef, but now would give a parse error at `b`, since the `a` would be disambiguated as an identifier declarator rather than another type specifier.

### DIFF
--- a/grammars/edu.umn.cs.melt.ableC/concretesyntax/DeclSpecifiers.sv
+++ b/grammars/edu.umn.cs.melt.ableC/concretesyntax/DeclSpecifiers.sv
@@ -50,6 +50,13 @@ autocopy attribute givenQualifiers :: ast:Qualifiers;
  -}
 synthesized attribute attributes :: ast:Attributes;
 
+{-
+ - This is needed to handle the ambiguity introduced by C11 repeated typedefs.
+ - In typedef int a; typedef int a; the second a must be disambiguated as an
+ - Identifier_t, not a TypeName_t, once we have already seen a type specifier.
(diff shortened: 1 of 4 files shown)


remexre (approved):
Works for the previously failing case (`#include <stdio.h>`, `#include <emscripten.h>`)

PR #173, opened 2020-11-20
Fix issues with anonymous struct type handling

krame505:
This fixes a bug with the vector extension reported by @remexre, giving internal errors for vectors of anonymous structs.  The fix turned out a bit more complicated than I originally thought, as without a tag name there is no way to look up and maintain a consistent refId.

### DIFF
--- a/grammars/edu.umn.cs.melt.ableC/abstractsyntax/host/Expr.sv
+++ b/grammars/edu.umn.cs.melt.ableC/abstractsyntax/host/Expr.sv
@@ -250,13 +250,13 @@ top::Expr ::= lhs::Expr  deref::Boolean  rhs::Name
   top.pp = parens(ppConcat([lhs.pp, text(if deref then "->" else "."), rhs.pp]));
   
   local isPointer::Boolean =
-    case lhs.typerep.withoutAttributes of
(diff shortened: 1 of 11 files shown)


grammars/edu.umn.cs.melt.ableC/abstractsyntax/host/Expr.sv
```diff
@@ -381,7 +381,7 @@ top::Expr ::= ty::TypeName  init::InitList
     | arrayType(_, _, _, _), _, _ -> []
     | t, nothing(), _ when init.maxIndex < 0 -> [err(top.location, s"Empty scalar initializer for type ${showType(t)}.")]
     -- Check that this type has a definition
-    | t, just(_), [] -> [err(top.location, s"${showType(t)} does not have a definition.")]
+    | t, just(_), [] -> [err(top.location, s"${showType(t)} ${hackUnparse(t)} does not have a definition.")]
```
remexre, line 384: Is this a debugging thing?

remexre (approved):

grammars/edu.umn.cs.melt.ableC/abstractsyntax/host/Expr.sv
krame505: Whoops, good catch!

PR #174, opened 2020-11-20
Fix Function Attributes

counc009:
Ran into this bug that the function attributes were pretty-printed in the wrong place on functionDecl.


krame505 (approved):

counc009:
@krame505 Is this a failure from this change? Looking at recent builds it looks like this may be a result of changes in ableC-algebraic-datatypes yesterday (that resulted in build failures yesterday).

krame505:
Hmm, that may be the case.  Currently we don't have extensions set up to build other extensions downstream.

krame505:
@remexre is fixing it right now, I will merge this once Jenkins passes.

PR #178, opened 2020-11-23
Fixes to typechecking related to array initializers

krame505:
Fixes some bugs that I encountered while working on tidying things up with the string extension.  Also a few more tests are passing now!


remexre (approved):
lgtm

PR #180, opened 2020-12-17
Fix Type-Checking Problems with _Atomic types

counc009:
This is attempting to fix errors in ableC that caused all assignments involving `_Atomic` types (atleast `_Atomic int`) to result in type errors. It seems that there were two issues, one that the atomicType production was not providing equations for isArithmeticType (and other similar attributes) based on the base type and that the `compatibleTypes` function only considered types compatible if both were atomic.

A simple example that failed previously but succeeds now:
```
int main() {
  _Atomic int x = 2;
  return 0;
}
```

This solution may not be ideal because it allows the `_Atomic` to be discarded easily and while GCC doesn't seem to complain about implicit casting of `int` to `_Atomic int`, it goes warn about implicit casting between `_Atomic int*` and `int*`. I don't know if we have access to the actual C11 standard, but if we do it might be worth checking how type-checking with atomics is supposed to work.


krame505 (approved):
Looks good to me!

PR #183, opened 2020-12-23
Refactors array declarations using new nonterm.

hypnoticOxbow:
Both DirectAbstractDeclarator_c and DirectDeclarator_c are effected.

New Nonterminal is called `PostFixModifier_c`, and it allows us
to extend the syntax of array declarations, allowing the addition
of productions like "checked" (for checkedc reasons) on that new nonterminal. An example of the new syntax that this should allow is below:
```
int a checked[10];
```
Also moves the definitions of DirectDeclarator_c, AbstractDeclarator_c,
and DirectAbstractDeclarator_c up to just after the definition of
Declarator_c, for readability.

### DIFF
--- a/grammars/edu.umn.cs.melt.ableC/concretesyntax/Declarations.sv
+++ b/grammars/edu.umn.cs.melt.ableC/concretesyntax/Declarations.sv
@@ -90,6 +90,73 @@ concrete productions top::Declarator_c
       top.ast = dd.ast; }
 
 
+closed nonterminal DirectDeclarator_c with location, declaredIdent, declaredParamIdents, ast<ast:TypeModifierExpr>, givenType;
+concrete productions top::DirectDeclarator_c
+| '(' d::Declarator_c ')'
+    { top.declaredIdent = d.declaredIdent;
+      top.declaredParamIdents = d.declaredParamIdents;
+      d.givenType = ast:parenTypeExpr(top.givenType);
+      top.ast = d.ast;
+    }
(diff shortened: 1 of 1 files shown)


grammars/edu.umn.cs.melt.ableC/concretesyntax/Declarations.sv
```diff
@@ -380,175 +447,78 @@ concrete productions top::Pointer_c
...
-    }
-| dd::DirectAbstractDeclarator_c
-    { dd.givenType = top.givenType;
-      top.ast = dd.ast;
-    }
-| p::Pointer_c
-    { p.givenType = top.givenType;
-      top.ast = p.ast;
-    }
+synthesized attribute exprAst :: ast:Expr;
```
krame505, line 450: What is this used for?

krame505 (commented):
Looks ok, besides the above.

grammars/edu.umn.cs.melt.ableC/concretesyntax/Declarations.sv
hypnoticOxbow: ah, thats an attribute that I forgot to remove. I'll get rid of it

hypnoticOxbow: Ok, this is now done with 1c4f92c

krame505 (approved):
Looks good!

PR #185, opened 2021-02-07
Add error checking for the placement of break and continue

### LINKED ISSUE #184: Missing Errors on break and continue
Currently we have don't produce errors if a break is placed outside of a loop/switch statement or if a continue is placed outside of a loop. I don't think this would be too hard to fix, we could introduce two new boolean attributes to just track whether a break/continue is valid and produce appropriate error messages when we encounter a break/continue when the associated attribute is false. 
These attributes probably gets placed on the same nonterminals as `returnType`, and they will generally just be copied from parent to child with loops setting them both to true and switch statements setting the attribute for break to true. (I'll try to check the C standard just to make sure there are no other caveats to their usage)

I may get around to this sometime during this semester, if not I put a reminder on my phone for May so I'll eventually get around to this.

counc009:
Resolves #184 (I got around to this sooner than I expected)

Introduces two new attributes `breakValid` and `continueValid` which are boolean and represent whether a break or continue statement, respectively, are valid in the current location in the tree. These are autocopy attributes and we set them to true in places in the tree where they become valid (a loop body or the body of a switch for a break).  These attributes occur on the same nonterminals as `returnType` because (as far as I can tell) there are pretty much no places a return is allowed where a break/continue isn't (as long as its within an appropriate construct). 
I made the errors more general than actually specifically saying that break/continue are only valid in a loop (or switch) since with extensions there may be more places they can be placed (or some types of loops where they aren't allowed). 

Also, this requires updates to a number of extensions; I've done my best to make sensible choices for these attribute values (mostly mimicking how `returnType` was handled), but it might be worth having someone at some point go through and figure out what we should really do, since this change is rather powerful--we can now prohibit break/continue statements from certain blocks of code which I imagine could be useful in the Halide extension for example.

Also, I don't think there's an issue for this already, but we need some way to get locations onto at least return/break/continue statements so that they can report sensible error locations (though would this be fixed by origin tracking?)

### DIFF
--- a/grammars/edu.umn.cs.melt.ableC/abstractsyntax/construction/FoldHelpers.sv
+++ b/grammars/edu.umn.cs.melt.ableC/abstractsyntax/construction/FoldHelpers.sv
@@ -73,7 +73,8 @@ Parameters ::= l::[ParameterDecl]
   return case l of
   -- A special case.  "type name(void)"  means no parameters.
   | [d] ->
-    case decorate d with {env = emptyEnv(); returnType = nothing(); position = 0;} of
+    case decorate d with {env = emptyEnv(); returnType = nothing(); position = 0; 
+            breakValid=false; continueValid=false;} of
       parameterDecl(nilStorageClass(), builtinTypeExpr(nilQualifier(), voidType()), baseTypeExpr(), nothingName(), nilAttribute()) -> nilParameters()
     | _ -> foldr(consParameters, nilParameters(), l)
     end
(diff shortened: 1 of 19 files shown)


krame505 (approved):
Looks good, although it's a bit annoying that this shows up in so many places.  Perhaps `returnType`, `breakValid`, and `continueValid` could be merged into a single attribute `controlStmtContext` (or something...needs a better name)?  

But this is also fine for now if you don't want to change everything again.

counc009:
Yeah, I like that idea, but that's a bit more of a project than I have time to tackle right now. I'll open an issue proposing that and I can self-assign and plan to work on it once I have more time.

I'll plan to merge this PR probably sometime this evening, if that's alright. Also, since I haven't merged into ableC recently, I don't remember if there's a procedure for merging the extensions?

PR #187, opened 2021-05-22
Control Statement Context Revision

### LINKED ISSUE #186: Merge returnType, breakValid, continueValid into one attribute
With the introduction of `breakValid` and `continueValid` attributes (#185) this increases the number of attributes we have to decorate expressions, statements, etc. with and it would probably be nicer and cleaner to merge these all into one attribute, like `controlStmtContext`. The nicest implementation might be for this to have a nonterminal type with an empty flow-type and attributes for `returnType`, `breakValid`, and `continueValid` (we could maybe even make it a closed nonterminal to allow other attributes to be added, though I don't know how useful that would really be).

I would also propose that if we do this it might be worth discussing whether we could/should move labels out of the environment and try to move them onto something like this since labels and gotos are control-flow mechanisms. Basically, if we put the list of labels on this attribute it would be easier to drop all labels inside some new construct (and therefore cause errors if code tries to use a `goto` to some label declared outside of the extension construct). This would allow extensions to introduce new constructs that could provide more guarantees about appropriate control-flow, which might be nice from a correctness perspective.

Self-assigning, and I'll try to get around to this towards the end of the semester.

counc009:
This resolves #186 by replacing the existing `returnType`, `breakValid`, and `continueValid` attributes with a `controlStmtContext` attribute whose type is a `ControlStmtContext` nonterminal which contains the three previous attributes, but now since they are one `Stmt` (and other constructs) as only a single attribute this does clean up code some since often the single attribute can be copied or modified in a single line, rather than having to update multiple attributes.

In addition, this moves labels out of the environment and onto this same attribute. This makes it easier to drop labels from visibility, since it just requires updating this attribute rather than messing with the environment. To achieve this, I do have to create an addition attribute `labelDefs` which tracks defined labels which are then propagated up the tree, much the way they were in `functionDefs`, but they are separated since `functionDefs` is related to the environment and the labels no longer are.

There are also changes in many of the ableC extensions, but I don't think there's anything too major there.


krame505 (approved):
Are `functionDefs` ever used for anything besides labels?  If not we should maybe just consider getting rid of that attribute.  

Otherwise, looks good - thanks.

counc009:
That's a good point. Doing a quick grep the only thing that appears to add elements to `functionDefs` now (at least in ableC itself) is in the `parameterDecl` production, which can add a `valueDef` into `functionDefs`. It seems like if that's the only use we may be able to refactor things and get rid of the attribute, but I'm not certain.

PR #199, opened 2022-10-24
Fix the character constant regexes to support the escapes in C11.

remexre:
Can we get this in before the release?

Also, what's the deal with needing `[\\]`? Without it, the regex doesn't match anything, but this feeeeeeels like a workaround for something awful.

(As an aside, the complexity of this regex is making me wish for a more verbose but more readable DSL we compile to terminal regexes...)


krame505 (approved):

krame505:
> Also, what's the deal with needing [\\]? Without it, the regex doesn't match anything, but this feeeeeeels like a workaround for something awful.

Do you mean as opposed to writing just `\\`?  That should work, AFAIK, if not then probably a bug?  The regexes were originally written back when there were some bugs with host escaping worked inside Silver regexes.

remexre:
> Do you mean as opposed to writing just `\\`?

Yeah, that's what I mean

PR #205, opened 2023-08-08
Fix bugs with struct field env and refIds

krame505:
Note that this is based on #204.


krame505:
Somehow this got created with the wrong base branch...

PR #217, opened 2024-10-22
Update for changes to decoration semantics in Silver

### LINKED ISSUE #208: Remove decExpr and friends
We should remove the `dec*` wrapper productions once all uses have been eliminated from ableC extensions.

krame505:
Depends on https://github.com/melt-umn/silver/pull/849.

Initial, mostly automated refactor pass to get ableC building with https://github.com/melt-umn/silver/pull/849.  Downstream extensions are disabled for now; they are being updated and re-enabled as a part of #218.


remexre (approved):
LGTM, conditional on #218 getting merged too

Issue #221, opened 2024-11-20
_FORTIFY_SOURCE=3 -related problems?


remexre:
```console
$ java -Xss6M -Xmx256M -jar compiler.jar examples/skeleton.xc -Iinclude
/nix/store/bl5b4ywrn1ls6fwc0cvsvwmr9hynr3gk-glibc-2.39-52-dev/include/bits/stdio2.h:31:6: error: Undeclared value __builtin_dynamic_object_size
/nix/store/bl5b4ywrn1ls6fwc0cvsvwmr9hynr3gk-glibc-2.39-52-dev/include/bits/stdio2.h:45:7: error: Undeclared value __builtin_dynamic_object_size
/nix/store/bl5b4ywrn1ls6fwc0cvsvwmr9hynr3gk-glibc-2.39-52-dev/include/bits/stdio2.h:45:54: error: argument 5 expected type void (got __gnuc_va_list)
/nix/store/bl5b4ywrn1ls6fwc0cvsvwmr9hynr3gk-glibc-2.39-52-dev/include/bits/stdio2.h:55:7: error: Undeclared value __builtin_dynamic_object_size
/nix/store/bl5b4ywrn1ls6fwc0cvsvwmr9hynr3gk-glibc-2.39-52-dev/include/bits/stdio2.h:69:8: error: Undeclared value __builtin_dynamic_object_size
/nix/store/bl5b4ywrn1ls6fwc0cvsvwmr9hynr3gk-glibc-2.39-52-dev/include/bits/stdio2.h:69:55: error: argument 6 expected type void (got __gnuc_va_list)
/nix/store/bl5b4ywrn1ls6fwc0cvsvwmr9hynr3gk-glibc-2.39-52-dev/include/bits/stdio2.h:198:14: error: Undeclared value __builtin_dynamic_object_size
/nix/store/bl5b4ywrn1ls6fwc0cvsvwmr9hynr3gk-glibc-2.39-52-dev/include/bits/stdio2.h:210:14: error: Undeclared value __builtin_dynamic_object_size

$ NIX_HARDENING_ENABLE="" java -Xss8M -jar compiler.jar examples/skeleton.xc -Iinclude

$ NIX_HARDENING_ENABLE="" java -Xss8M -jar compiler.jar examples/skeleton.xc -D_FORTIFY_SOURCE=3 -Iinclude -O2
[...]

PR #222, opened 2024-11-20
Implements __builtin(_dynamic)_object_size, describes reasons for ignoring builtins

remexre:
Depends on https://github.com/melt-umn/silver/pull/862.

@krame505 , do you remember what the `indexOf("t", x.lexeme) != -1 || indexOf("u", x.lexeme) != -1` condition was about?

### DIFF
--- a/grammars/edu.umn.cs.melt.ableC/abstractsyntax/host/Overload.sv
+++ b/grammars/edu.umn.cs.melt.ableC/abstractsyntax/host/Overload.sv
@@ -16,8 +16,6 @@ top::Expr ::= @fn::Expr args::Exprs result::Expr
 
 synthesized attribute callProd::Maybe<Call> occurs on Type, ExtType;
 flowtype callProd {decorate} on Type, ExtType;
-synthesized attribute callWithUnevaluatedArgsProd::Maybe<Call> occurs on Type, ExtType;
-flowtype callWithUnevaluatedArgsProd {decorate} on Type, ExtType;
 
 dispatch MemberAccess = Expr ::= @e::Expr deref::Boolean name::Name;
 
@@ -401,7 +399,6 @@ top::Type ::=
 {
   top.arraySubscriptProd = nothing();
   top.callProd = nothing();
-  top.callWithUnevaluatedArgsProd = nothing();
   top.memberCallProd = nothing();
   top.memberProd = nothing();
   top.exprInitProd = nothing();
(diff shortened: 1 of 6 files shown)


remexre:
oops, just reread the code and realized we need `t`. Is there a reason we need the `u` test, though?

krame505 (approved):
Seems fine to me, as long as nothing breaks.  I don't remember what t and u stand for in these specs.

remexre:
- `t` = this has custom `t`ypechecking rules (in the first dozen I saw, typically it's generic)
- `u` = this builtin's arguments are `u`nevaluated at runtime -- it's only used in 5 builtins, but one of them is `__builtin_dynamic_object_size`, as in #221; the other 4 are:
  - `__builtin_constant_p` (also has `t`)
  - `__builtin_classify_type` (also has `t`)
  - `__builtin_object_size`
  - `__builtin_os_log_format_buffer_size` (also has `t`)

I guess the question is whether we do any analysis that assumes arguments _are_ evaluated for effect at runtime -- if so, I might:

- tweak the environment to store builtin-ness in the types, since at least GCC disallows taking a function pointer to a builtin anyways, so we should have a check for that
- make an AST-only prod for "call to a builtin whose arguments aren't evaluated"
- make calls to these builtins and `sizeof` use that AST prod

?

krame505:
Yeah, that's probably the right fix.

grammars/edu.umn.cs.melt.ableC/abstractsyntax/host/Overload.sv
```diff
@@ -16,6 +16,8 @@ top::Expr ::= @fn::Expr args::Exprs result::Expr
 
 synthesized attribute callProd::Maybe<Call> occurs on Type, ExtType;
 flowtype callProd {decorate} on Type, ExtType;
+synthesized attribute callWithUnevaluatedArgsProd::Maybe<Call> occurs on Type, ExtType;
```
krame505: Wait, why is this dispatching based on type? I thought this would be dispatching on looking up a name?

remexre: yeah that was just me copy-pasting the Call implementation; the current one doesn't actually dispatch. more notes on that below

remexre:
- currently, this just adds a production for the builtins w/ unevaluated args (BWUA), which does _not_ have any sort of dispatching
- this is mainly because I don't want to have to have one production per BWUA; right now it'd only be two, but still, unless we need it I'd rather not make people write extra equations
- i _would_ be interested in making a BWUA nonterminal, instead of using `Name` as the `func` NT
- ...except type-checking checks argument types on the argument, using the type from the function
- i think we should do it the other way around: `Exprs` produces a list of types, those get passed around to the FunctionType, it does the comparison
- this way, we can handle polymorphic builtins too
- i can do this in this PR, but i'd just as rather not till we need that

PR #228, opened 2024-11-23
Fix to locking for extension make invocations

krame505:
This ensures that we avoid invoking silver again to build an extension artifact after aquiring the lock, if the artifact is up-to-date.  This fixes a race condtion where no-op silver builds for up-to-date grammars were still modifying the jars, causing concurrent builds depending on those jars to fail.

### DIFF
--- a/extension.mk
+++ b/extension.mk
@@ -146,31 +146,49 @@ generated lib:
 
 # Locking is used to ensure that multiple parallel make invocations in the same workspace
 # don't lead to race conditions causing a dependency to be built more than once.
-ifeq ($(MAKELEVEL),0)
 LOCK=flock $(dir $@)
-else
-LOCK=
-endif
 
 $(ABLEC_JAR): $(shell find $(ABLEC_BASE)/grammars/ -name *.sv -print0 | xargs -0)
+ifeq ($(MAKELEVEL),0)
 	$(LOCK) $(MAKE) -C $(ABLEC_BASE) ableC.jar
+else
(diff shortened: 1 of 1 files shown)


extension.mk
```diff
@@ -146,31 +146,49 @@ generated lib:
...
 # Note that $(DEP_JARS) are order-only dependencies, to avoid expensive rebuilds.
 # If dependency extension syntax changes, this may require `make depclean` to be reflected.
-silver-compiler.jar: $(wildcard grammars/*/artifacts/silver_compiler/*.sv) | $(DEP_JARS) generated
-	$(LOCK) silver -o $@ $(SVFLAGS) $(EXT_GRAMMAR):artifacts:silver_compiler
+$(SV_COMPILER_JAR): $(wildcard grammars/*/artifacts/silver_compiler/*.sv) | $(DEP_JARS) generated
+  ifeq ($(MAKELEVEL),0)
+	$(LOCK) $(MAKE) $@
+  else
+	silver -o $@ $(SVFLAGS) $(EXT_GRAMMAR):artifacts:silver_compiler
+  endif
```
remexre, line 175:
```suggestion
ifeq ($(MAKELEVEL),0)
	$(LOCK) $(MAKE) $@
else
	silver -o $@ $(SVFLAGS) $(EXT_GRAMMAR):artifacts:silver_compiler
endif
```

for consistency

remexre (approved):

PR #235, opened 2025-04-02
Avoid using the extended Silver compiler for linking

krame505:
See https://github.com/melt-umn/silver/pull/893.


unironically (approved):
Reviewed in MELT meeting today.

